PR #4, opened 2016-02-05
rhine: Fix wifi with Enforcing SELinux

alviteri:
Avoid

03-01 18:50:37.286 W/init    (1): type=1400 audit(0.0:4): avc: denied { write } for name=wcnss_wlan dev=tmpfs ino=11360 scontext=u:r:init:s0 tcontext=u:object_r:wcnss_device:s0 tclass=chr_file permissive=0


alviteri:
needed for QCOM wifi as rhine yukon and kanuti

PR #10, opened 2016-02-14
common: Add handheld_core_hardware permission XML

galaxyfreak:
When this is missing, custom ROMs break in unpredictable ways, with all launchers crashing, and certain screens in the Settings app crashing, in what looks like OOM errors. This is probably true on AOSP too, but less obvious.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -38,7 +38,8 @@ PRODUCT_COPY_FILES += \
     frameworks/native/data/etc/android.hardware.bluetooth_le.xml:system/etc/permissions/android.hardware.bluetooth_le.xml \
     frameworks/native/data/etc/android.hardware.nfc.xml:system/etc/permissions/android.hardware.nfc.xml \
     frameworks/native/data/etc/com.android.nfc_extras.xml:system/etc/permissions/com.android.nfc_extras.xml \
-    frameworks/native/data/etc/android.software.midi.xml:system/etc/permissions/android.software.midi.xml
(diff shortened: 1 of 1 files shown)


AdFad666:
Oh boy this took me all week to figure out, such a non-obvious bug! I was trying to track down the OOM causes playing with ART configs etc, but I eventually had to go back and redo the entire process of platform/device -> common/platform/device to find what got lost.

jerpelea:
it is already defined 
https://android.googlesource.com/platform/build/+/master/target/product/full_base_telephony.mk#28

AdFad666:
Not everywhere https://github.com/CyanogenMod/android_build/blob/cm-13.0/target/product/full_base_telephony.mk

IMO it is a good idea to include it here because of the non obvious way it breaks things - nothing in the logs tells you this permission is missing, you just get OOM crashes.

Many people will struggle to find why their not AOSP ROM is crashing.

AdFad666:
CM made this change, but many things are based on CM, I see the same change in many places

https://github.com/CarbonROM/android_build/blob/lp5.1.1/target/product/full_base_telephony.mk
https://github.com/SlimRoms/android_build/blob/mm6.0/target/product/full_base_telephony.mk

jerpelea:
change should not be made here since it is intended for devices with GSM

AdFad666:
OK I now see why it can be an issue here. Will you accept it per device, similar to the tablet line mentioned above?

jerpelea:
since problem is in cm and other derivates i think that the best place to add it would be cm_device.mk

PR #22, opened 2016-02-29
common: Fix tad_static denials

humberos:
This patch removes duplicate tad_static domain and
uses tad domain instead of it.

Also, it makes tad as a permissive domain since it requires
specific block_device permissions that are labeled as
neverallow android rule.

Change-Id: If88fde7ac13736fcd714891fe61eda4757b77116


humberos:
I am currently using selinux as enforcing mode on karin.

PR #30, opened 2016-03-29
Update init.common.rc

jvquaglio:
When using full disk encryption with a password different from DEFAULT_PASSWORD, the class 'main' is started to prompt for password/pin/pattern and then it is reset. The services qmuxd and netmgrd are declared as class 'main' and 'disabled'. Because they are disabled, they need to be explicitly called to be started. When the class 'main' is reset, these services never go up again. Changing the 'start' to 'enable' solves the issue...

### DIFF
--- a/rootdir/init.common.rc
+++ b/rootdir/init.common.rc
@@ -390,8 +390,8 @@ on property:ro.boot.baseband=apq
     stop ril-daemon
 
 on property:ro.boot.baseband=msm
-    start qmuxd
-    start netmgrd
+    enable qmuxd
+    enable netmgrd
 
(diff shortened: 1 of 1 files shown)


rootdir/init.common.rc
```diff
@@ -390,8 +390,8 @@ on property:ro.boot.baseband=apq
     stop ril-daemon
 
 on property:ro.boot.baseband=msm
-    start qmuxd
-    start netmgrd
+    enable qmuxd
+    enable netmgrd
 
 on property:persist.radio.multisim.config=dsds
     start ril-daemon2
```
jerpelea: the same change is needed here

PR #32, opened 2016-04-17
sony: common: sepolicy: Fix tad denials

humberos:
It requires external/sepolicy changes.
Add tad as a sepolicy attribute and fix all permissions.

Signed-off-by: Humberto Borba <humberos@gmail.com>
Change-Id: I991b66671bc8d2bec4df3f4a69df75ad02c46318

PR #33, opened 2016-04-17
sony: common: Set SELinux as enforcing status

humberos:
Change-Id: Id500aaf78997566159f5317338f671c6cb20bb71


humberos:
it requires https://github.com/sonyxperiadev/device-sony-common/pull/32

jerpelea:
Open devices is intended for development
Selinux should not be set as enforcing!

PR #40, opened 2016-05-04
sony: common: Add common FMRadio.mk setup

humberos:
If device supports FMRadio then it should include this file.

Change-Id: I81a107a6b3df9a18701f625a1048a55a3f017522


humberos:
if device supports then it should include common config.
That is the idea. ;)
https://github.com/sonyxperiadev/device-sony-rhine/pull/206

PR #44, opened 2016-05-06
DO NOT MERGE: sepolicy: rebuild entire sepolicy

AdFad666:
The original policy was taken from CM. It is incomplete, it has some
dangerous policies, and depends on QCOM sepolicy. This is a complete
rebuild of the sepolicy taking only the needed parts from the QCOM
repo.


AdFad666:
Zero denials on ivy on first boot, with wifi connected.

Next steps:
* harder testing, GSM, bluetooth, gapps, etc etc
* compare to nexus sepolicy to remove any qcom wtfbbq

PR #46, opened 2016-05-12
Revert "sony-common: add ADBoverWIFI prop"

zhuoyang:
This commit make adb over wifi persist over reboot , even when disabled in developers option.
Not needed for average users, can be enabled by editing build.prop.


AdFad666:
I'm against this change, I use it regularly.

These repos are mainly for development so this is a useful feature, if you want to create builds for average users then you should probably fork and make any changes suitable for average users (you should also change selinux to enforcing).

PR #50, opened 2016-05-16
sepolicy: rebuild entire sepolicy

AdFad666:
The original policy was taken from CM. It is incomplete, it has some
dangerous policies, and depends on QCOM sepolicy. This is a complete
rebuild of the sepolicy taking only the needed parts from the QCOM
repo, and tightening up the bad parts.

According to Google, an unnecessarily large policy:

* Takes a double hit on memory as the policy sits in the
  ramdisk and is also loaded into kernel memory.
* Wastes disk space by necessitating a larger bootimage.
* Affects runtime policy lookup times.

https://source.android.com/security/selinux/device-policy.html

This policy has had some testing:

* Pure AOSP Master branch with GApps
* WiFi, Bluetooth, NFC connections
* GSM connections, data and calling
* YouTube video playback over bluetooth headphones

Right now there are a couple of denials due to Google's ongoing
refactoring efforts, and there will probably be a few more after
wider testing, but this policy is at least 95% complete.

Tested on these devices:

* suzuran
* ivy
* karin_windy
* sirius
* honami

Signed-off-by: Adam Farden <adam@farden.cz>


AdFad666:
Zero denials on ivy on first boot, with wifi connected.

Next steps:
* harder testing, GSM, bluetooth, gapps, etc etc
* compare to nexus sepolicy to remove any qcom wtfbbq

AdFad666:
I still want to test on:
* honami
* leo
* karin_windy
* suzuran

PR #58, opened 2016-05-25
Add Kitakami fingerprint HAL

BuzzBumbleBee:
* This was based around the OnePlus Two TrustZone fingerprint HAL (found here
   https://github.com/CyanogenMod/android_device_oneplus_oneplus2/blob/cm-13.0/fingerprint)

  * Extra information was obtained from the Nexus 6P (angler kernel driver) (found here
   https://android.googlesource.com/kernel/msm/+/3d0a564505ab8452e7e6f52b972db386cc2f5f69)

  * Protocol informations was verified as it passed through the QSEECOM API to linux kerne

  Authors Jens Andersen & Shane Francis

### DIFF
--- a/fingerprint/Android.mk
+++ b/fingerprint/Android.mk
@@ -1,0 +1,35 @@
+#
+# Copyright (C) 2016 Shane Francis / Jens Andersen
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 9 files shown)


EnJens:
Depends on https://github.com/sonyxperiadev/device-sony-common/pull/59 and https://github.com/sonyxperiadev/device-sony-kitakami/pull/178

AdFad666:
When building on AOSP Master:
```
device/sony/common/fingerprint/fingerprint.c:29:10: error: incompatible pointer to integer conversion initializing 'uint64_t' (aka 'unsigned long long') with an expression of type 'void *' [-Werror,-Wint-conversion]
uint64_t challenge = NULL;
         ^           ~~~~
```

PR #78, opened 2016-07-09
sony-common: fingerprint: allow suzu

alviteri:
allow suzu to build common fingerprint HAL


jerpelea:
suzu uses a different implementation

PR #85, opened 2016-08-21
Enable camera hal version 3

mkdmc:
I have compared camera stability between camera hal v3 and old hal. Camera hal v3 is much more stable. When I set persist.camera.HAL3.enabled to 0, I can only take few pictures(maybe 5 pics) and then "can't connect to camera" message shows and I should restart phone to make camera working again. But when I enable hal v3, I can take many pictures without camera connectivity error.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -254,7 +254,7 @@ PRODUCT_PROPERTY_OVERRIDES += \
 # Camera
 PRODUCT_PROPERTY_OVERRIDES += \
     camera.disable_zsl_mode=0 \
-    persist.camera.HAL3.enabled=0 \
+    persist.camera.HAL3.enabled=1 \
     persist.camera.gyro.disable=1 \
     persist.camera.feature.cac=0 \
     persist.camera.ois.disable=0


alviteri:
https://github.com/sonyxperiadev/device-sony-common/commit/43c867123ce8915326a13b951e7f6421adb6a868

alviteri:
Also if you are testing with a rhine device then you are pushing to wrong branch..... Because master is used for 1.3.3 kernel

mkdmc:
I have pushed to right branch.
I know that camera has more resolutions and more iso values and other camera apps also work when using hal v1, but it's useless when we can only take few pictures. So I think hal v3 should be as default hal.

xboxfanj:
Some applications seem to function better on devices such as the OnePlus 3 with HAL1.  For that, I suggest the following commits: https://gerrit.aospa.co/#/c/190/ https://gerrit.aospa.co/#/c/191

mkdmc:
It seems both hal 1 and hal 3 are now relatively stable.(I tested on 2016-08-23)

PR #86, opened 2016-08-23
[Nougat] Updates

alviteri:


olefb:
Looks like it switched to https://android.googlesource.com/platform/external/dhcpcd-6.8.2/+/android-7.0.0_r1 according to the manifest: https://android.googlesource.com/platform/manifest/+/b299c955daf3ebd3a8d016263b5c0027d960fa0c%5E%21/#F0

alviteri:
You can see any trace of dhcp.conf in that repo?

olefb:
There is a dhcpcd.conf there, but I see that [the AOSP devices](https://android.googlesource.com/device/huawei/angler/+/874f26e293be999fb6a566f4e1ef37fc16537517%5E%21/#F0) have made the same change, so ignore my post. :-)

AdFad666:
Nexus devices have it removed from init files too

PR #93, opened 2016-08-30
Finally unbreak dualsim :)

Quarx2k:
Fix typo.

### DIFF
--- a/rootdir/init.common.rc
+++ b/rootdir/init.common.rc
@@ -259,7 +259,7 @@ service ril-daemon2 /system/bin/rild -c 2
     class main
     socket rild2 stream 660 root radio
     socket sap_uim_socket2 stream 660 bluetooth bluetooth
-    socket rild2-debug stream 660 radio system
+    socket rild-debug2 stream 660 radio system
     user root
     disabled
     group radio cache inet misc audio log


Myself5:
@Quarx2k I'd be intrested in what is needed for the dualsim devices to work.
Do you also require some blob related changes in comparision to a non dsds phone or is using your commit and the proper defconfig to append the proper dtb (so that sdcard is working too) enough?

Quarx2k:
Yes sdcard works fine. Everything works except 5ghz wifi.
You need this commit, ds dtb and this blob https://github.com/Quarx2k/device-sony-ivy_dsds/blob/cm-13.0/temp/rild
Tommorrow i check what wrong with AOSP rild and hopefully make a patch, but for now only rild blob is solution.

zachariasmaladroit:
@Quarx2k where did that rild originate from ? please let us know what is wrong with the original rild,

Great work ! :+1:

Quarx2k:
rild from sony stock 6.0.1 ds rom.
I debugged aosp rild, there one problem. 
RILD1 connects to qmi RIL[0] socket (it's ok), but RILD2 connects also to qmi RIL[0], but should to RIL[1]. Thats why ril-daemon2 constantly crashing and IMEI for sim2 not available.
Probably should be easy to fix, will check tomorrow.

Myself5:
The 5GHz WIFI is something @AdFad666 might be able answer. He said throwing out the SIM card and reinserting it fixes 5GHz WIFI (its still something which has to get solved properly).

Also wouldn't there be any commits be needed for AOSP for general DualSIM Support? Like Software/UI Wise or does all of that work?

Cause if everything works with AOSP, thats something @jerpelea might be intrested in, in reference to our last dsds discussions :P

zachariasmaladroit:
From my knowledge the Dual SIM support should be there starting with Lollipop,

CyanogenMod builds further on that,

not really sure if the Kernel-side needs further modifications but I also ported over the SIM_DETECT driver and it seemed to have helped (as did Quarx2k),

I don't own a dual sim device and also didn't get any meaningful logcat, dmesg, etc. output up until now - so it's kind of hard to debug

Quarx2k:
I think SIM_DETECT driver not really needed. I added it to debug simcards on kernel level.
In stock rom it used for SemcSimDetection, which reboot device when simcard removed/inserted.

AdFad666:
Yes, for me on my ivy, if you:

1. disable WiFi
2. insert a SIM
3. reboot the device
4. enable WiFi

you will be able to connect to 5GHz WiFi. This but also affects karin_windy butis impossible to solve in the same way due to lack of RIL.

Fortunately for ivy this fix persists across reboots and power downs, even when the SIM has been removed.

Quarx2k:
AOSP rild works fine, no need any patches, just need change socket name :)

alviteri:
welcome to this family @Quarx2k xD

PR #95, opened 2016-08-31
common: enable FD in video

alviteri:


jerpelea:
fd is not yet enabled in the vendor

PR #96, opened 2016-09-08
cmdline: only enable permissive and debug on eng builds

AdFad666:
security security security!

Signed-off-by: Adam Farden <adam@farden.cz>


tomgus1:
+1

Myself5:
+1 Security is love

jerpelea:
this is a develpment base and we will keep the selinux permisisve

PR #98, opened 2016-09-15
Add broadcom fm radio stuff

Quarx2k:
Needed for Broadcom BT/FM combo chip kernel driver.
Allow use FMRadio on devices with broadcom stack.

brcm-uim-sysfs from sony opensource package.
FMRadio-jni from https://android-review.googlesource.com/#/c/46100/
libfmradio.v4l2-fm, Author: marco.sinigaglia@csr.com


jerpelea:
this commit has no commit message and no indication from where the code comes

PR #103, opened 2016-09-17
sony-common: remove ADBoverWIFI prop

alviteri:
will be replaced by ExtendedSettings


jerpelea:
on the eng builds device will not reach ui

PR #104, opened 2016-09-17
lights: Add 12-bit backlight compatibility

kholk:
Add a check for 12-bit backlight using max_brightness sysfs file
while retaining compatibility with 8-bit.

I could have added 0-32bit backlight compatibility but that would
have uselessly increased computational complexity to the library
since we actually use only 8-bit or 12-bit backlight in our
devices.


kholk:
Needs testing on Ivy, Satsuki and other 12-bit backlight devices.
Also testing on regular 8-bit backlight devices is preferred.

kholk:
Tested working on Suzuran. Needs test on old 8-bit backlight devices.

kholk:
Testing done. Ready to be merged.

PR #129, opened 2016-10-22
Enable night display by default

rcstar6696:
It will not use HWC2.0 it will be done via SW

Will only work on android 7.1


alviteri:
you know sw means more drain on battery life?

rcstar6696:
Not really much i tested it a day 
But we should enable the possibilty of night mode 
Its much better to read on dark rooms

PR #143, opened 2016-11-25
Setup seccomp policy

BuzzBumbleBee:
7.1 now makes use of minijail on kernel versions newer than 3.8, this seccomp policy allows mediaserver / meidacodecs access to the needed syscalls


BuzzBumbleBee:
Tested with dora on LA.UM.5.5.r1 kernel branch, no more mediacodec crashes & ringtone samples play fine

PR #179, opened 2016-12-23
sony-common: enable vendor specific camera format

alviteri:
neeed for a patch in displayHAL to avoid video play lags in Nougat with AOSP camera


alviteri:
https://android-review.googlesource.com/#/c/316755/

PR #182, opened 2017-01-02
sony: common: Tunning swappiness

humberos:
This control is used to define how aggressive the kernel will
swap memory pages throught zRAM.

Up to 100
Default is 60

the logic of this patch is to force the kernel to addresses the memory pages
to zram block available (compressed memory way), consequently,
this increases the amount available memory to system.


humberos:
ping @AdFad666

jerpelea:
swapiness is too agresive and vill create memory pressure

PR #187, opened 2017-01-11
Sony-common: build gps location from repo

alviteri:
Do not use prebuilt libs anymore


rancidfrog:
@alviteri I get compile error when I try to build from repo: https://github.com/sonyxperiadev/vendor-qcom-opensource-location/issues/3
Am I missing something?

PR #199, opened 2017-01-24
sony-common: healthd: add sony custom healthd

alviteri:


alviteri:
requires https://android-review.googlesource.com/#/c/327458/

PR #227, opened 2017-03-07
common: Rename & set rfkilldisabled property for brcm_fmradio.

Diewi:
Rename rfkilldisabled to rfkilldisableduim in the parsing code
of UIM such that we can set rfkilldisabled in the build.prop to
stop android's bt implementation to mess around with the chip's
power states.
This allows to enable BT while the FMRadio is on.

Change-Id: I6639e21ee4c1ca592ed70f417c825ed2a987be71
Signed-off-by: Alexander Diewald <Diewi@diewald-net.com>

### DIFF
--- a/brcm_fmradio/brcm-uim-sysfs/upio.c
+++ b/brcm_fmradio/brcm-uim-sysfs/upio.c
@@ -93,7 +93,7 @@ static int is_rfkill_disabled(void)
 {
     char value[PROPERTY_VALUE_MAX];
 
-    property_get("ro.rfkilldisabled", value, "0");
+    property_get("ro.rfkilldisabled.uim", value, "0");
     UPIODBG("is_rfkill_disabled ? [%s]", value);
 
     if (strcmp(value, "1") == 0) {
--- a/common.mk
+++ b/common.mk
@@ -304,3 +304,9 @@ PRODUCT_PROPERTY_OVERRIDES += \
 # sdcardFS
 PRODUCT_PROPERTY_OVERRIDES += \
     ro.sys.sdcardfs=true
+
+# BT/FMRadio
+ifeq ($(filter rhine kanuti,$(SOMC_PLATFORM)),)
+PRODUCT_PROPERTY_OVERRIDES += \
+    ro.rfkilldisabled=1
+endif


Diewi:
As discussed earlier, this commit solves the following bug:
1. Turn off BT
2. Start the FMRadio
3. Turn on BT while the FMRadio is playing

--> Observe that BT does not turn on any more and will not again until reboot. Also FMRadio stops playback.

alviteri:
is like ugly conditional's struct xD anyway good work @Diewi

ifeq ($(BOARD_HAVE_BCM_FM), true)
PRODUCT_PROPERTY_OVERRIDES += \
           ro.rfkilldisabled=1
endif

better ?

humberos:
LGTM!
Good job!

alviteri:
@Diewi good job .. but are not many spaces into that conditional ?

alviteri:
@Diewi I mean something like this xD
https://hastebin.com/araqahixak.pl

alviteri:
also ... since the conditional contains a prop it can be moved to common.mk and not CommonConfig.mk

humberos:
It will not work @alviteri 
just because BCM_FM flag (chicken x egg)

alviteri:
@humberos did you remember cameraHAL prop ? for 8974 it worked under conditional

Diewi:
@alviteri It is done in CommonConfig.mk, because the flag is not available, when common.mk is evaluated. The if was placed there at first, butit did not work out.

humberos:
@alviteri 
cameraHAL was using SOMC_PLATFORM which is declared in platform.mk
https://github.com/sonyxperiadev/device-sony-loire/blob/master/platform.mk#L22

The BCM_FM is declared in "PlatformConfig"
https://github.com/sonyxperiadev/device-sony-loire/blob/master/PlatformConfig.mk#L86

That's the reason why cameraHAL works and FM will not

alviteri:
Can be excuse me xD

Diewi:
Set the rfkilldisabled property based on the flag SOMC_PLATFORM: The flag BOARD_HAVE_BCM_FM is not available in common.mk & setting  PRODUCT_PROPERTY_OVERRIDES in CommonConfig.mk gets overridden (it is not called from device.mk).
Also, implement the filter to exclude the platforms rhine and kanuti since newer platfrom will have a Broadcom chip. Thanks @alviteri for the valuable suggestions!
I made a mistake checking this property in the generated build.prop, hence I did not notice it was not set by the initial implementation.

PR #228, opened 2017-03-07
common: Rename & set rfkilldisabled property for brcm_fmradio.

Diewi:
Rename rfkilldisabled to rfkilldisableduim in the parsing code
of UIM such that we can set rfkilldisabled in the build.prop to
stop android's bt implementation to mess around with the chip's
power states.
This allows to enable BT while the FMRadio is on.

Change-Id: I6639e21ee4c1ca592ed70f417c825ed2a987be71
Signed-off-by: Alexander Diewald <Diewi@diewald-net.com>


jerpelea:
rebased

PR #244, opened 2017-03-27
[master] Override android file system config

humberos:
Override android file system config and allow derived projects
to use their particular implementation.

Also merge kitakami, loire and tone sec_config to match with CAF UID.


humberos:
tested on kagura!
it works!

PR #251, opened 2017-05-06
Revert "enable fluence"

alviteri:
This should be set per platform and not common

This reverts commit afbffe4.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -239,9 +239,7 @@ PRODUCT_PROPERTY_OVERRIDES += \
     vidc.debug.level=1
 
 # Audio
-# Fluencetype can be "fluence" or "fluencepro" or "none"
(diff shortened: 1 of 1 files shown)


rancidfrog:
Wouldn't changes be needed to the following lines as well:
```
persist.audio.fluence.voicecall=true \
    persist.audio.fluence.voicecomm=true \
    persist.audio.fluence.voicerec=false \
    persist.audio.fluence.speaker=true
```
Then, they should all be set to false?
Or, does it not matter regardless?

Still no news on working mic incall¿

alviteri:
what's the problem with mic incall ? for me it is working 8952

rancidfrog:
8974 has incall echo and volume issues.
Also when you use speakerphone, they cannot hear you properly, like some sounds are chopped

PR #255, opened 2017-05-14
common: maximize dex optimization

tetsuo55:
added all options except for ab
our devices have no ab partitions

disabled profiling on some to prevent following errors that cause a
bootloop:
Suppressed: java.lang.IllegalStateException: everything-profile is profile-guided, but not allowed for shared-apk
Suppressed: java.lang.IllegalStateException: everything-profile is profile-guided, but not allowed for forced-dexopt

Change-Id: I64b028d941855b6d7b60d57058fb438627f555f3

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -173,6 +173,19 @@ PRODUCT_COPY_FILES += \
     device/sample/etc/old-apns-conf.xml:system/etc/old-apns-conf.xml \
     device/sample/etc/apns-full-conf.xml:system/etc/apns-conf.xml
 
+# Select the highest setting of dex optimization
+PRODUCT_PROPERTY_OVERRIDES += \
+    dalvik.vm.dex2oat-filter=everything-profile \
+    dalvik.vm.image-dex2oat-filter=everything-profile \
+    pm.dexopt.bg-dexopt=everything-profile \
+    pm.dexopt.boot=everything-profile \
+    pm.dexopt.core-app=everything-profile \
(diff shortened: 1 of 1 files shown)


kholk:
Tested on very old device that is not covered by Open Devices.

Needs to be tested on Open Devices supported platforms.

tetsuo55:
Also i recommend the removal of dexpropt after this is merged.
The profile guided optimization will create files in data anyway making the preopt somewhat useless

kholk:
DEXPREOPT gives us a very nice first boot performance boost.
Moreover, we have plenty of space on /system partition and the pre-optimization makes it possible to use more space in /system and less in /data.

erikcas:
I completely agree with @kholk

pablomh:
We don't have that much space in /system in tulip...

When I added the userland binaries to support exFAT in my tree I run out of space.

kholk:
Take Tulip as an exception.... it's basically the ONLY device that has a smaller system partition, so perhaps you can disable it on your tree, or only on Tulip.

Whatever, this PR's subject is not DEXPREOPT, so..

daedroza:
Any final decision this?

jerpelea:
here is a good reference howto set it 
https://source.android.com/devices/tech/dalvik/configure

jerpelea:
i would preffer to follow AOSP defaults and let every custom ROM choose its own optimisation

PR #258, opened 2017-05-18
[COMMON] Implement new RQBalance-PowerHAL and librqbalance

kholk:
This is a new PowerHAL, made specifically for RQBalance, which
is our first choice hotplugging, running on all Sony devices
on Open Devices kernel.

The new PowerHAL includes possibility to get advanced interaction
with the RQBalance driver in kernel by using a supplementary
library allowing this.
In case no supplementary library is provided, the functionality
will be restricted to basic governor parameter setting.

This HAL supports various power modes applicable to RQBalance:
BATTERYSAVE - Low performance mode, mostly to save battery
BALANCED - Sweet spot for good performance and battery
PERFORMANCE - Maximum performance, but increased heat and battery
              consumption.
OMXDECODE - Video HW decoding
OMXENCODE - Video HW encoding

The PowerHAL PowerServer is responsible for
communicating with external libraries (for now, the only user
is librqbalance) to get custom optimization parameters.
The PowerHAL PowerServer, an UNIX sockets server implementation,
was made because of a quirk in Android, which makes IPC between
32<->64 bit library practically impossible in standard ways.

Timed optimizations implement POSIX timers in the HAL, responsible
for setting up a timer after which the specified optimization
(mincores, maxcores, balance level, frequency tweaks etc) will
be discarded and reverted to sane platform defaults.
This is necessary, for example, when we want to set specific
optimizations that will make the SoC to draw much more power
in very busy situations: here we want to minimize the IPC and
wasted cpu cycles to calculate.
A good example would be this:
1. User starts video encoding through camcorder
2. Camcorder needs to allocate buffers and initiate hw encoder
3. Camcorder will start recording

On step 2, we may want to push the SoC busses to the MAX to get
maximum bandwidth on both RAM and SDHCI controller, which will
make the camcorder application to allocate the required buffers
and the file on SDCard/eMMC in a very small amount of time.
While pushing up the frequencies will draw a lot of power,
finishing the task in WAY LESS time will make us save time AND
eventually also a small amount of power (computation at low
power for 1 second would draw more power compared to high power
for 50msec).
Of course, the technique applies to many more situations.

About configuration, the Android properties required for full funcitionality of the HAL
were getting to be way too many and anyway, this way has never
really been clean at all: since the beginning, at least 10 props
were needed to get *at least some* functionality.

Now we are about to implement more hints and more power modes for
the HAL to be more interactive on more events and, for each new
power mode, we get 5 new properties. This is unacceptable.

Implement configuration parsing via an XML file, located on the FS
at /system/etc/rqbalance_config.xml.

The configuration XML file has to adhere to the following format:
`<rqbalance_config>`
`    <POWER_MODE_STRING>`
`        <cpuquiet PROPERTY="VALUE"/>`
`        <rqbalance PROPERTY="VALUE"/>`
`    </POWER_MODE_STRING>`
`</rqbalance_config>`

------ librqbalance -------
About librqbalance: This library implements perf_lock_acq and perf_lock_rel, called
by various Qualcomm HALs.
The library is made specifically for Sony devices on Open Devices
kernel, featuring rqbalance as hotplugging governor.
It interacts with the PowerHAL's PowerServer, sending specific
(timed or not) optimizations on-demand, so this library will
work ONLY with the new PowerHAL featuring PowerServer, working
through UNIX sockets.


power/power.c
```diff
@@ -18,168 +18,542 @@
...
+        case POWER_MODE_BALANCED:
+            type_string = "balanced";
+            compat_string = "normal";
+            break;
+        case POWER_MODE_PERFORMANCE:
+            type_string = "performance";
+            compat_string = "perf";
+            break;
+        case POWER_MODE_OMXDECODE:
+            type_string = "video_decoding"
```
erikcas: type_string = "video_decoding";

power/power.c
```diff
@@ -18,168 +18,542 @@
...
+            compat_string = "vdec";
+            break;
+        case POWER_MODE_OMXENCODE:
+            type_string = "video_encoding"
```
erikcas: missing ";"??

type_string = "video_encoding";

power/power.c
```diff
@@ -18,168 +18,542 @@
...
+
+    return ret;
+}
+
+/*
+ * power_init - Initializes the PowerHAL structs and configurations
+ */
+static void power_init(struct power_module *module UNUSED)
+{
+    int ret, dbg_lvl;
```
erikcas: Add i to int?

int i, ret, dbg_lvl;

see line 438

kholk:
I'm sorry, during the cleanup phase some typo occurred because one commit got lost in the squash.

Fixed.

erikcas:
thanks
also, is librqbalance built by default or should we add to common.mk:

# librqbalance
PRODUCT_PACKAGES += \
    librqbalance

kholk:
Yes it's correct. I'm also adding the required property. Hold on one sec.

kholk:
There you are.

hahnjo:
This code is not yet in this repository's `n-mr1` but the related pull requests have been merged to `n-mr1` for the device repositories - is that correct?

kholk:
Thanks for pointing that out. No, it wasn't correct, the PR for n-mr1 has been created.

PR #266, opened 2017-05-27
[k4.4 - READ COMMENT] init.common.usb.rc: Complete rewrite for USB ConfigFS gadget

kholk:
Kernel 4.4 supports ONLY ConfigFS (so no FunctionFS) and it is
enabled by default, hence we need to configure it accordingly.

Please note that this change can be valid also on kernel 3.18,
where ConfigFS USB gadget can be activated, but FunctionFS is
enabled by default.

### DIFF
--- a/rootdir/init.common.usb.rc
+++ b/rootdir/init.common.usb.rc
@@ -11,102 +11,143 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-on init
-    write /sys/class/android_usb/android0/iSerial ${ro.serialno}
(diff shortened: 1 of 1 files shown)


alviteri:
@kholk just an idea why not let older and this ifdef by kernel 4.4 only ?

kholk:
We can leave both configurations, there is a property that you activate when you want android to use configfs..... But this is master branch. 

Master branch is for latest and greatest things.... And 4.4 is THE bleeding edge.
If you are worried about 3.10 kernel devices.... Don't worry (I am evaluating the risks and benefits of porting configfs on 3.10) and remember that this is master branch.

kholk:
PR updated with major cleanup of all functions, removal of QC proprietary debug functions and addition of MIDI, audio_source, accessory, found in the latest version(s) of Android.

PR #274, opened 2017-06-28
[URGENT] [COMMON] Ensure directory creation and correct perms for PowerServer

kholk:
RQBalance-PowerHAL PowerServer will create it automatically
but, for some reason, on some Android versions the directory
will be created with wrong permissions.
Fix this by paranoidly ensuring directory creation and by doing
permission changes at init time.

IMPORTANT NOTE:
Do NOT use "mkdir powerhal 0773 system system": as I wrote,
the directory may be already there at that time!!

### DIFF
--- a/rootdir/init.common.rc
+++ b/rootdir/init.common.rc
@@ -110,6 +110,11 @@ on post-fs-data
 
     # FM Radio
     mkdir /data/misc/fm 0770 system system
+    
+    # RQBalance-PowerHAL PowerServer
+    mkdir /data/misc/powerhal
+    chmod 0773 /data/misc/powerhal
+    chown system system /data/misc/powerhal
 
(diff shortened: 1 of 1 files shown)


kholk:
Urgent. Merge me ASAP.

rcstar6696:
Lol i just posted in the slack group sometime before that the permission is not correct but i got ignored 

This should fix powerserver Start

kholk:
I was examining this thing since the last week.... it happens only in certain conditions, I don't actually really know the real reason.
I'm posting this only now because of two things:
1. Christian actually remembered me of the existance of init.rc (lol seriously, didn't even think of this)
2. I was actually looking for a possible race condition that was producing the problem.

N.2 wasn't effective, so N.1 was the accepted way :P

PR #277, opened 2017-07-08
[master] sony: common: overlay: Add ambient display (notifications)

humberos:
Change-Id: I99fbff020cbc7e5a446959ec10999a6bbf82dd38


humberos:
although it works fine but it is not compatible with dt2w.

humberos:
dt2w is still broken even after clearpad driver update which means that we have some work on incell panel driver. So ambient display is ok to go.

PR #279, opened 2017-07-15
common: Add MiFare permission file

Myself5:
Change-Id: Ic8b0f75766945ec12cd090a5d1f74ac08b78c08a


alviteri:
Make attention with this an Rhine

Myself5:
Rhines PN544 does not have MiFare support, moving the copy lines to the platform trees.

PR #285, opened 2017-08-04
rootdir: init.common.usb.rc: remove unneeded mkdir call

oshmoun:
The folder /config is already created at this point in the boot process, so no need to do that again.
Not only that, but the permissions declared in the mkdir call are wrong for Android.
Android can then not write to /config, and as a result cannot configure /config/sdcardfs.

Change-Id: I001f73adcebb550748ce18fc40b69033ba1e5aa4


Myself5:
MERGE MERGE MEEEERRRRGEEEEEE

Myself5:
Oh and: Merge please :D

PR #291, opened 2017-08-16
sony-common: remove android_filesystem_config

alviteri:
actually it is needed by qmux and used by loire only .. remove this from common


jerpelea:
please do not remove it

PR #297, opened 2017-08-25
use /charger symlink to launch healthd

alviteri:
Makes this consistent with other targets and maken sure
any changes in healthd will be transparent to targets as long
as the symlink is maintained.

Update the seclabel accordingly too.

Test: Tested charge-only mode on angler

Change-Id: Ic71e32f52c39bc45eae8d2552259695934592fc7
Signed-off-by: Sandeep Patil <sspatil@google.com>


alviteri:
@jerpelea this + https://github.com/SonyAosp/platform_system_core/commit/f7195a22117d1a5725697fe32f16c6567e0185a4 fixed offline_charging mode

PR #307, opened 2017-09-15
sony-common: add missing hal services

alviteri:


jerpelea:
please clean the commit!
It contains changss that are not needed

PR #315, opened 2017-09-21
sony-common: power: move powerHAL to /vendor

alviteri:
under /system/lib/hw - /system/lib64/hw it give this
E RQBalance-PowerHAL: Error writing to /sys/devices/system/cpu/cpuquiet/nr_power_max_cpus: Invalid argument

having maple on powersave mode
maple:/ # cat sys/devices/system/cpu/cpuquiet/nr_power_max_cpus
3

under /vendor/lib/hw - /vendor/lib64/hw maple can run on performance mode
maple:/ # cat sys/devices/system/cpu/cpuquiet/nr_power_max_cpus
8


alviteri:
ping @kholk

kholk:
I agree but we can't do that.......... We need to find an alternative solution to this.

alviteri:
Ok

PR #316, opened 2017-09-22
rqbalance: move socket path out of /data

oshmoun:
Having the socket for PowerServer under /data caused issues with
encryption where cryptfs would try to unmount /data, but would fail
because the power hal would still hold its socket open.
Move the socket over to /dev/socket/rqbalance/ to avoid such issues, and
to also keep it together with other sockets.


kholk:
/dev/socket/powerhal/rqbsvr

PR #318, opened 2017-09-25
sony-common: drop net_raw group

alviteri:
Use CAP_NET_BIND_SERVICE instead
following marlin


jerpelea:
it has to be tested on all devices

alviteri:
Maple and Dora ... Will build later on kugo

alviteri:
kugo ok here

PR #319, opened 2017-09-28
sony-common: remove unneeded HDMI symlink

alviteri:
on aosp we are not using HDMI also on legacy platforms it was symlinked to FB1 if you see at
msm8956-mdss.dtsi
msm8996-mdss.dtsi
msm8998-mdss.dtsi

the new platforms are using hdmi under FB2 not FB1


alviteri:
ping @kholk

PR #337, opened 2017-10-19
Fix incorrect condition checking

vladimir-zahradnik:
Android.mk has defined several conditional expressions which check for absense of explicitly listed platforms and based on that perform some actions. However except first conditional, all other expressions are defined inside first expression, e.g.:

```make
ifneq ($(filter yukon rhine shinano kanuti kitakami loire,$(PRODUCT_PLATFORM)),)
...
ifneq ($(filter loire tone yoshino,$(PRODUCT_PLATFORM)),)
endif
endif
```

So for instance if I was building for ```loire``` platform, check for it would be evaluated at least twice and when I was building ```rhine``` platform, no code would be executed at all.

Following commit moves all conditional expressions to same global level, e.g.:
```make
ifneq ($(filter yukon rhine shinano kanuti kitakami loire,$(PRODUCT_PLATFORM)),)
...
endif

ifneq ($(filter loire tone yoshino,$(PRODUCT_PLATFORM)),)
...
endif
```

If this behavior is an error and not an intention, please merge this PR.

Thanks!

Signed-off-by: Vladimir Zahradnik <vladimir.zahradnik@gmail.com>

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -4,6 +4,8 @@ LOCAL_PATH := $(call my-dir)
 
 include $(call all-makefiles-under,$(LOCAL_PATH))
 
+endif
+
 ifneq ($(filter shinano kitakami,$(PRODUCT_PLATFORM)),)
 $(shell mkdir -p $(PRODUCT_OUT)/system/vendor/firmware)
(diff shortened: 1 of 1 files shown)


jerpelea:
initial condifion checks if we have a sony device suported by project 
the second one will add some simlinks only for some platfroms included in the first condition

PR #344, opened 2017-10-20
Enable Ambient Display

Bitti09:
Tested on kagura@oreo with  telegram  only at the moment. 

Needs more testing on other devices.

Dreams are disabled by default with this commit.

### DIFF
--- a/overlay/frameworks/base/core/res/res/values/config.xml
+++ b/overlay/frameworks/base/core/res/res/values/config.xml
@@ -207,4 +207,15 @@
          The default is false. -->
     <bool name="config_lidControlsSleep">true</bool>
 
+    <!-- Enable doze mode
+         ComponentName of a dream to show whenever the system would otherwise have gone to sleep. -->
+    <string name="config_dozeComponent">com.android.systemui/com.android.systemui.doze.DozeService</string>
+
+    <!-- If true, the doze component is not started until after the screen has been turned off
+         and the screen off animation has been performed. -->
+    <bool name="config_dozeAfterScreenOff">true</bool>
+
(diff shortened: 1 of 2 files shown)


alviteri:
unfortunately we have not AMOLED displays

Bitti09:
So its useless for recent sony devices?
Kernel says that the panel is completly off so there shouldn't  much extra  battery usage

alviteri:
I think we should create a doze service as some Lineage devs did for IPS devices ...... I tested this in 3.10 kernel but was not merged for some reasons

kholk:
What uses the most current in our phones is the display backlight. Whatever, if this commit makes a disabled by default setting to appear for the user to choose to eventually enable the feature, I'm in.

Bitti09:
it only makes the settings appear and dreams disabled but still visible. currently no random reboots while sleeping on my kagura.

Bitti09:
@kholk it's not ready to merge. I need to fix some more settings to get it ready for merging

kholk:
Fine, I repeat... as long as Ambient Display is not enabled by default, I'm okay with having it.

Bitti09:
currently its enabled by default. i'm testing an commit that sets the default state to disabled

kholk:
Perfect!

Bitti09:
required commit for default settings needs to be  in fw/base so i don't think that the commit  here will get merged.
Anyways: if anyone wants it  here  is it:
https://github.com/SlimBit-AOSP/framework_base/commit/4f0c81a2d5c82fd0bffb69ec3c40bd5afc0ac5b4

PR #349, opened 2017-10-21
sony-common: add recovery headers

alviteri:
needed for inline kernel system build
this needs extra commit to call this from bootable/recovery/minui/Android.mk
https://github.com/SonyAosp/platform_bootable_recovery/commit/8f6832b3ad16bdad4af298315e8fcfb3f6b41fc8

### DIFF
--- a/CommonConfig.mk
+++ b/CommonConfig.mk
@@ -85,6 +85,9 @@ ifeq ($(HOST_OS),linux)
 endif
 DONT_DEXPREOPT_PREBUILTS := true
 
+# Recovery Header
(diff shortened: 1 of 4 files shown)


alviteri:
maple poplar dora recovery works

alviteri:
ping @kholk

Bitti09:
are we doing inline kernel build on default sod settings?
if  i remember right  the prebuilt one is used by default .

alviteri:
This is not needed here .. but is usefully for all external projects basing ROMs for Sony Xperia on this source

jerpelea:
Headers should be properly exported.There is no meed for a local copy

PR #351, opened 2017-10-24
[N-MR1] Rename odm build properties file with standard filename

abioteau:
FS config (on 8.0 and master) define specific file access rights only for /odm/build.prop and /odm/default.prop


jerpelea:
the build properties should not include the odm properties file
this file is used by the version checking tool only

PR #358, opened 2017-11-14
[master] Use z3fold as zRAM backend

ab123321:
We added z3fold to kernel, but it isn't used by default.
This change makes use of this driver.

Requires https://github.com/sonyxperiadev/kernel/pull/1515


jerpelea:
NO NOT EXTEND THE CMDLINE !

PR #365, opened 2017-12-11
Initial odmcheck (1/4)

magnus4:
Odmcheck will block init until exited. If a mismatch between odm partition and the build props is detected a warning screen will be shown. Currently this warning screen just has a simple placeholder UI with the version information shown on a blue background. The warning screen will be shown for 10 seconds, then the boot will resume regardless. Once everything is in place and works fine, after changing a build flag, odmcheck will instead shutdown the device after the 10 seconds.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -214,6 +214,10 @@ PRODUCT_PACKAGES += \
 endif
 endif
 
+# odmcheck
+PRODUCT_PACKAGES += \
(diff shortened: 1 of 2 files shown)


jerpelea:
please squash the commits.

magnus4:
Squashed

PR #373, opened 2017-12-16
[master] sony: common: Move fstab mount process to common

humberos:
Also, do not call swap mount until completed boot process
This help to save ~50ms in boot time and bootanim shown time

Reference:
https://android.googlesource.com/device/google/wahoo/+/06b4b0ce139eea0e015184297cedbbbc9cf2d8d3

Change-Id: Ic63fb53689d7da8ca948cedfd4f41d9869f64033

### DIFF
--- a/rootdir/init.common.rc
+++ b/rootdir/init.common.rc
@@ -54,6 +54,8 @@ on charger
     start irsc_util
 
 on fs
+    mount_all ./fstab.${ro.boot.hardware}
+
     # Start HW service manager early
     start hwservicemanager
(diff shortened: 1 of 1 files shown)


humberos:
@jerpelea 
If you agree then I will patch platforms.

Also, we need to work on this change 
https://github.com/sonyxperiadev/device-sony-common/commit/1c9a06052c858f31ad4653113ff0bfea685d299a
since this breaks tone boot process.

Ping @oshmoun

PR #374, opened 2017-12-17
[master] Revert "init: speedup the boot process"

humberos:
This reverts commit 1c9a06052c858f31ad4653113ff0bfea685d299a.

Fix boot process for tone platform.

Change-Id: Ia9d75f969cdc33b182449fe12ff2a6490c672c7d

### DIFF
--- a/rootdir/init.common.rc
+++ b/rootdir/init.common.rc
@@ -56,35 +56,10 @@ on charger
 on fs
     mount_all ./fstab.${ro.boot.hardware}
 
-    # Start HW service manager early
-    start hwservicemanager
-
-    # qseecomd needs /dev/block/bootdevice
-    # vold needs keymaster that needs qseecomd
-    start qseecomd
-
-on late-fs
-    # Start services for bootanim
-    start surfaceflinger
-    start bootanim
-    start hwcomposer-2-1
-    start configstore-hal-1-0
-    start gralloc-2-0
-
-on post-fs
-    # set RLIMIT_MEMLOCK to 64MB
-    setrlimit 8 67108864 67108864
-
-    # Wait qseecomd started
-    wait_for_prop sys.listeners.registered true
-
 on post-fs
     exec - system graphics -- /vendor/bin/odmcheck
 
 on post-fs-data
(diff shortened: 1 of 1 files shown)


humberos:
Please hold on, I will split this patch locally and try to figure out what service is breaking tone boot process

jerpelea:
please update your device to the latest sw with Emma and use latest vendor
I tried to replicate the bug and devices boot normal

humberos:
Device: Kagura
SW: 41.3.A.0.401
ODM: V6

@oshmoun also have the same issue on keyaki.

oshmoun:
Yes can confirm,and I also on the latest fw (because i can flash odm through fastboot). Somehow tone doesn't like it when display services are started too early.

humberos:
Thanks @oshmoun 
I will close this PR soon since it is only a reference to fix tone boot process.
I will test one by one of display services and see if there is a property that should be tested before the call statement.

bartcubbins:
it's ok on suzu (k4.4) and tulip(k3.10) on 8.1.0_r1 with hwc2.0

humberos:
Yeah, suzu and yoshino are ok!
Problem is on tone family.
Thanks for the confirmation @bartcubbins

bartcubbins:
very strange

humberos:
Ok!
I've found the issue.
It is custom bootanimation...

@jerpelea 
I am sorry. You're right.
There is no boot problem when I use default android bootanimation.

oshmoun:
but a custom bootanim doesn't cause the same issue on loire (omni as an example)

humberos:
yeah! too fast
I've got the bootloop again even by removing custom bootanimation

humberos:
ccrap.... WTH??

oshmoun:
your guess is as good as mine :D 
maybe it's worth it to try out with hwc2
https://github.com/sonyxperiadev/device-sony-common/pull/375

humberos:
The problem is... 
    start bootanim

humberos:
https://github.com/sonyxperiadev/device-sony-common/pull/376

PR #381, opened 2017-12-18
[master] Cleanup

bartcubbins:


bartcubbins:
@jerpelea I think maybe left one ifeq for k4.4 on master branch?

PR #385, opened 2017-12-19
Drop "mkdir /data/data/media"

jamuir:
this command prevents the encryption policy from being set correctly
on the directory `/data/data` when a device first boots in FBE mode.


jamuir:
@jerpelea: do you know what `mkdir /data/data/media ...` was required for?  Is the path correct?

jerpelea:
can be safely removed

PR #387, opened 2017-12-19
[o-mr0] Move fingerprint common stuff

bartcubbins:
TARGET_PER_MGR_ENABLED and TARGET_FPC_VERSION flags can be placed in common repo because all our current devices require these flags and fingerprint overlay and they are these are the same for all platforms.


jerpelea:
please do not move the settings overlay for fingerprint location

PR #395, opened 2017-12-22
[master] CommonConfig: Enable TARGET_USES_COLOR_METADATA

oshmoun:
Enable the feature to handle the color metadata in display hal.
Otherwise, hal complains about an unknown parameter type:
03-18 22:14:09.559   501   640 E DisplayMetaData: Unknown paramType 2

Change-Id: Ia185b6497dfed269c8723241f2f33fdffe458eb4


oshmoun:
to be used together with https://github.com/sonyxperiadev/device-sony-common/pull/375

oshmoun:
definitely
tested here on kugo
needs testing on tone and yoshino devices

PR #399, opened 2017-12-31
[DONOTMERGE] gps: switch to an updated hal

oshmoun:
The msm8994 gps hal we are using is obsolete. Switch to using a more up-to-date hal.

Change-Id: I4449e66f772c298a040d71cf8d3aca14b290aa07


oshmoun:
msm8998 hal might also be valid, if updated oss location is compatible. Please do not merge until that is tested by @humberos

PR #400, opened 2018-01-01
[master] rootdir: init: services: add xtra-daemon

oshmoun:
This daemon is required by the sdm845 gps hal to update AGPS data.

Change-Id: Ie3dd95d45030a8a8cabb27d1f800262f1b68310e


oshmoun:
included in #401

PR #401, opened 2018-01-02
[master] New GPS fixes

humberos:
1. Update gps.conf from wahoo
2. Add xtra daemon as service
3. Add system group to xtra since GPS Hal (system_server) will write location sock_file.

There is progress with GPS.


humberos:
it requires
https://github.com/sonyxperiadev/device-sony-sepolicy/pull/310

humberos:
ping @oshmoun 
feel free to change your patch (adding system group) since PR is still opened.
gps is working good here on kagura

oshmoun:
This is working very well on kugo

PR #402, opened 2018-01-02
[master] sony: common: init: Do not call bootanim services early

humberos:
It is still breaking boot process for tone platform.

Change-Id: Iad230f13a9f14d4d03472f06af670cf5dfc507c7


humberos:
ping @oshmoun

PR #403, opened 2018-01-02
[o-mr0] New GPS

humberos:
Add new GPS from master.


humberos:
ops.... there is no sdm865 at this version

jerpelea:
Hups cherry pick it😉

PR #407, opened 2018-01-06
[DONOTMERGE] rootdir: system: gps_debug: add SUPL server config

oshmoun:
The gps hal does not seem to read SUPL data from gps.conf
Hence, add the configuration to gps_debug.conf to allow android to set those

Change-Id: Ib14d5f782b9b6095360186582a514e83dff69b80

### DIFF
--- a/rootdir/system/etc/gps_debug.conf
+++ b/rootdir/system/etc/gps_debug.conf
@@ -6,8 +6,8 @@
 ##### AGPS server settings #####
 ################################
 # FOR SUPL SUPPORT, set the following
-# SUPL_HOST=supl.google.com or IP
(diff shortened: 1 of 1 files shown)


oshmoun:
I'd rather receive confirmation on this before it got merged.
So, here's how to confirm an improvement:
1. Install the apps "GPS Check" by Chris Brossard and "GPS Status & Toolbox" (any other gps apps that display ephemeris/almanac data  for each satellite and can force agps data updates are fine too)
2. Check for how many satellites there are ephemeris data available
3. Apply this PR
4. Reboot and force an AGPS update
5. Check again how many satellites have ephemeris data. The number should now be higher than in step 2

PR #414, opened 2018-01-14
manifest: add gnss service

oshmoun:
gnss service is built already, so add it in manifest to explicitly declare usage.

Change-Id: Ic39cf0f6edd636c6c30c1e4e9b1b571ea705ffc7

### DIFF
--- a/manifest.xml
+++ b/manifest.xml
@@ -76,6 +76,15 @@
         </interface>
     </hal>
     <hal format="hidl">
+        <name>android.hardware.gnss</name>
+        <transport>hwbinder</transport>
+        <version>1.0</version>
+        <interface>
+            <name>IGnss</name>
+            <instance>default</instance>
+        </interface>
+    </hal>
+    <hal format="hidl">
         <name>android.hardware.graphics.allocator</name>
         <transport>hwbinder</transport>
         <version>2.0</version>


oshmoun:
This requires https://github.com/sonyxperiadev/device-sony-sepolicy/pull/327

humberos:
seems this patch breaks systemui on kagura.
yes, I've added sepolicy patch.

humberos:
W zygote64: Waited one second for android.hardware.gnss@1.0::IGnss/default. Waiting another...

oshmoun:
Weird
It's running fine here on both kugo and kagura
Can you confirm that the service has the correct context?

humberos:
Hmmmm
Aosp or omni?

oshmoun:
Tested on omni so far

oshmoun:
Could you please check dmesg, to see if the service is not starting due to missing an selinux domain?

humberos:
[    3.941270] init: service gnss_service does not have a SELinux domain defined

oshmoun:
did you just build bootimage? That would explain this. Could you run restorecon and see if the domain of the service changes?

humberos:
Sorry @oshmoun 
The problem is on my side due to dirty builds.

oshmoun:
No problem :) 
good to know it works!

PR #415, opened 2018-01-22
8.1 bringup

alviteri:

PR #421, opened 2018-01-30
[master] sony: common: Enable vulkan compute feature

humberos:
This code change ensures that vulkan compute xml
is copied to /vendor/etc/permission which implictly
enable compute on vulkan

Change-Id: I4c9e1e7abb23baf74026237315068f42373b1d00


jerpelea:
david already pushed the same patch
please cherrypick
https://github.com/sonyxperiadev/device-sony-common/pull/415

humberos:
@jerpelea

Oh sorry
But we also have authorship preserved here (at least from qcom@6.4 tag )
But I will change the patch

jerpelea:
both look ok but you have to sync with david since he pushed first

jerpelea:
i agree that the commit message in this one is cleaner

alviteri:
@humberos this one was picked from wahoo

PR #422, opened 2018-01-31
[n-mr1] common.rc: Use setgid on /data/misc/wifi

abioteau:
The macaddrsetup service creates a file in this directory containing
the wifi MAC address that needs to be readable by the wireless
service. Ensure files created here are in the wifi group.


abioteau:
https://github.com/sonyxperiadev/device-sony-yoshino/blob/n-mr1/rootdir/init.yoshino.rc#L35

PR #424, opened 2018-02-01
Audio effect

richardtullberg:
@oshmoun 
These fixes might solve your BT issues.

### DIFF
--- a/rootdir/vendor/etc/audio_effects.xml
+++ b/rootdir/vendor/etc/audio_effects.xml
@@ -48,7 +48,7 @@
         <effect name="downmix" library="downmix" uuid="93f04452-e4fe-41cc-91f9-e475b6d1d69f"/>
         <effect name="loudness_enhancer" library="loudness_enhancer" uuid="fa415329-2034-4bea-b5dc-5b381c8d1e2c"/>
         <effect name="aec" library="qcom_pre_processing" uuid="0f8d0d2a-59e5-45fe-b6e4-248c8a799109"/>
-        <effect name="ns" library="qcom_pre_processing" uuid="72d30d20-0306-11e6-9860-0002a5d5c51b"/>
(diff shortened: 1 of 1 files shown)


luk1337:
Do they actually load?

luk1337:
https://github.com/sonyxperiadev/device-sony-common/blob/master/hardware/voice_processing/voice_processing_descriptors.c#L28
https://github.com/sonyxperiadev/device-sony-common/blob/master/hardware/voice_processing/voice_processing_descriptors.c#L40

luk1337:
tl;dr I took a look at AOSP audio HAL. It seems to load `/vendor/lib/soundfx/libqcomvoiceprocessingdescriptors.so` or `/system/lib/soundfx/libqcomvoiceprocessingdescriptors.so` depending on your setup and that's where it gets UUIDs from. So if this commit fixes your BT issues then it means that somehow your voice processing is broken and disabling it is the workaround for it cause obviously it wouldn't load it properly with suggested by @richardtullberg UUIDs.

luk1337:
His changes would work tho if you decided to drop custom `libqcomvoiceprocessingdescriptors` library.

jerpelea:
@richardtullberg which path do we take ?

richardtullberg:
Let's just leave it be if it works.

jerpelea:
@richardtullberg the problem is that on n-mr1 we use this patch and on o we closed it
we should test all Android versions and choose 1 direction 
now it is confusing !

luk1337:
I have alternative patch for that ~ https://gist.github.com/1a96bff6a6d99cddc0fda6baea78e3b6

luk1337:
It addresses everything I mentioned here.

jerpelea:
@richardtullberg?

richardtullberg:
Let's do it like @luk1337 does it then

jerpelea:
@luk1337 please pr on the following branches:

n-mr1-legacy 
n-mr1
o-mr0
master

PR #425, opened 2018-02-02
8.1

alviteri:


oshmoun:
Please also update the built services in common-treble.mk
for example:
https://github.com/sonyxperiadev/device-sony-common/blob/master/common-treble.mk#L41

alviteri:
Not needed read the code @oshmoun

oshmoun:
can't see how updating the manifest entries for the hals to a newer version will actually make a difference without also building the updated services.
Besides, it's a bit weird to have the same service with two different versions declared in the same repository.

alviteri:
As I said check the code https://android.googlesource.com/platform/hardware/interfaces/+/android-8.1.0_r4/wifi/1.1/default/Android.mk

alviteri:
That's the reason of XML to exists ..... You can and will control from there

Myself5:
The XML file is responsible for telling the system what HALs to load. If those HALs don't get built, you won't be able to load them either. So please use common sense and read  the code ( https://android.googlesource.com/platform/hardware/interfaces/+/android-8.1.0_r4/wifi/1.1/Android.bp#46 ), then update the Makefile to build the 1.1 HALs.

alviteri:
Another one ... Please "build" them and tell me if they got builded .....

alviteri:
Please send me a screen with your 1.1 rc service built

alviteri:
If that exist

AdrianDC:
On a general point, could you please Edit your comments instead of adding new minimal ones to avoid random emails? Thanks.

alviteri:
https://android.googlesource.com/device/google/wahoo/+/android-8.1.0_r4/device.mk#415
https://android.googlesource.com/device/google/wahoo/+/android-8.1.0_r4/manifest.xml#334

For better explanation to you

alviteri:
https://android.googlesource.com/platform/hardware/interfaces/+/android-8.1.0_r4/wifi/Android.bp as you can see the sub are called 1.0 and 1.1 you should tell by manifest which one use .... I hope is clear now

Myself5:
Most of the Services you just referenced are all wahoo specific, and NONE of the Increments were "only" 1.1 changes in the manifest. There is this handy feature called BLAME on git, maybe make yourself familiar with it before getting cocky towards us who wanted to provide constructive feedback.

PowerHAL: They introduced a 1.1 SERVICE to increment the manifest: https://github.com/CarbonROM/android_device_google_wahoo/commit/f3e845ce21adf40b4caa8982447b0bf99eeeee81
ThermalHAL: Changed name and service version in the makefile: https://github.com/CarbonROM/android_device_google_wahoo/commit/de912858f80ca2b1fbd380d0baefaf12fffecfde
VibratorHAL: Changed name and service version in makefile: https://github.com/CarbonROM/android_device_google_wahoo/commit/e2ac78d27f9e66b44b2c22d7c18b581e3a2ab025
USBHAL: Updated Version and service version in makefile:
https://github.com/CarbonROM/android_device_google_wahoo/commit/fc32e28be93531094932b66be5d55546283cda11 and https://github.com/CarbonROM/android_device_google_wahoo/commit/a0752793a5570f604368861c16861d49ffa40814

You are indeed right about WiFi, but you're wrong about all other HALs.
Power, Thermal, Vibrator and USB all lack a properly defined 1.1 Service, therefore won't build any 1.1 libs either, making your commit invalid. Wahoo has them because Wahoo has its own HALs.

alviteri:
please next time pick your DORA because you have sony device and make the proper test ... thanks

Myself5:
do you even read what I wrote? I never talked about the SELinux changes. I purely referred to the incremented version. So maybe stop trying to proof me with Pixel 2 commits which you clearly don't understand (which I proved above).

And if favour of stopping the email spam. Think about what you intend to write before you write it and put everything into a single message.

alviteri:
Just remove the useless service as you called in the comment and try rebooting device tell me see if it will boot ... To be more fast

Myself5:
Are you going crazy now? I never said to remove them either. I said there is no proper 1.1 Implementation for those HALs above, the code that exists is only used for device sided HALs, like the ones I pointed out, making the increment commit useless.

Ah whatever, you won't understand it anyways... 

**I -1 this PR. And would prefer it not to get merged the way it is.**

Thats my last statement on this PR.

alviteri:
ok Mr. tell me how enable 1.1 impl because for me they are in they only be switched from manifest ..... in the case we create own then they will be called 1.1-sony as wahoo did ... for now we can use android ones

Myself5:
It's worthless talking to you if you refuse to read what I said, David.

*There is no 1.1 impl, there is no 1.1 service using the 1.1 impl and there is no other way to us the 1.1 impl than using custom device services, which is not needed for our devices, for anything other than wifi.*

You should start seeing that these devices are not a Pixel 2 and should stop blindly copying every little diff you find.

jerpelea:
https://android.googlesource.com/platform/hardware/interfaces/+/dbd83ef5855a40840278fa865d23b31cc911cd2e

PR #428, opened 2018-02-16
common.rc: Create netmgrd dirs on /data in post-fs-data not of boot trigger

nexus511:
The boot trigger gets executed before the filesystems are ready. If an
encrypted filesystem is used, it is not yet mounted at the time of execution of
the boot trigger. This causes problems with "forceencrypt" settings in the
fstab, as the partition is also encrypted on the first boot and directories
can never be created at all.

sonyxperiadev/bug_tracker#78


jerpelea:
please update your patch

PR #429, opened 2018-02-18
rootdir: vendor: add public.libraries.txt

oshmoun:
Android needs this text file to know what libraries are allowed to be loaded by apps directly.
For now, add:
libOpenCL.so to the list, in order to get OpenCL to function properly.
adsprpc related libraries, since they were also declared public in stock.

Change-Id: I15ef7a7ab7d1f1a0cc6b484cd2ca9993f40de23b


uglavnom-bezopasni:
Commit cause hdr+ hang at processing,  To resolve delete public.libraries.txt and reboot.

PR #430, opened 2018-02-27
[master] sony: common: init: Fix charger dependencies

humberos:
This path allows system and odm to be mounted early which
have .rc files and critical services that are called/required
from boot process.

There is a race condition during offline charger service call as following:

When its dependencies are declared in vendor then we have the service loading error:

[    2.189134] charger: android charging is enabled
[    2.202449] EXT4-fs (mmcblk0p52): mounted filesystem with ordered data mode. Opts: barrier=1
[    2.207436] EXT4-fs (mmcblk0p28): mounted filesystem with ordered data mode. Opts: barrier=1
[    2.207589] init: do_start: Service rmt_storage not found
[    2.207648] init: do_start: Service mlog_qmi_service not found
[    2.207745] init: do_start: Service irsc_util not found
[    2.207809] init: processing action (charger) from (init.loire.pwr.rc:30)

If the services are declared inline (in ramdisk) then we have:

[    2.662462] charger: android charging is enabled
[    2.676562] EXT4-fs (mmcblk0p52): mounted filesystem with ordered data mode. Opts: barrier=1
[    2.684344] EXT4-fs (mmcblk0p28): mounted filesystem with ordered data mode. Opts: barrier=1
[    2.685980] init: starting service 'rmt_storage'...
[    2.686785] init: starting service 'mlog_qmi_service'...
[    2.687525] init: starting service 'irsc_util'...
[    2.688257] init: processing action (charger) from (init.loire.pwr.rc:30)

Change-Id: I927a23aaec1847b5ab8dfc8fda6172ce930d2c6e

### DIFF
--- a/rootdir/init.common.rc
+++ b/rootdir/init.common.rc
@@ -13,11 +13,6 @@
 # limitations under the License.
 
 on charger
-    # Booting modem
-    wait /dev/block/bootdevice/by-name/system
-    mount ext4 /dev/block/bootdevice/by-name/system /system ro barrier=1
-    wait /dev/block/bootdevice/by-name/oem
-    mount ext4 /dev/block/bootdevice/by-name/oem /odm ro barrier=1
     start rmt_storage
     start mlog_qmi_service
     start irsc_util
@@ -36,6 +31,9 @@ on early-init
 on init
     wait /dev/block/platform/soc/${ro.boot.bootdevice}
     symlink /dev/block/platform/soc/${ro.boot.bootdevice} /dev/block/bootdevice
+
+    # Mount RO partitions early and load all init files from /system/vendor
+    mount_all ./fstab.${ro.hardware} --early
 
     # Enable subsystem restart
     write /sys/module/subsystem_restart/parameters/enable_ramdumps 0
@@ -61,8 +59,6 @@ on init
     write /proc/sys/vm/page-cluster 0
 
 on fs
-    mount_all ./fstab.${ro.hardware}
-
     # /dsp is initially unlabelled so we need to mount
     # it as rw, restore AOSP labels, then remount
     restorecon_recursive /dsp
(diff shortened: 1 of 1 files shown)


humberos:
@jerpelea 
I think may we should move back all services that are critical to ramdisk.

btw, about this change, shall I create a new rc file with services ? (I mean like we did before)

@oshmoun 
I think may the tone problem (reboots during initialization) is related with critical services.
I mean, system nor odm are not ready when the service is called then device reboots. Its is a theory.

jerpelea:
instead of adding the services in init we should check why system is not mounted early

oshmoun:
The same services are found in loire's odm. Why is this issue not happening there?
Maybe finding that out will lead to a solution.

humberos:
@jerpelea 
patch updated

jerpelea:
https://github.com/sonyxperiadev/kernel/commit/58705ca63229eaa2bd04af12dd504c959f924543
https://github.com/sonyxperiadev/kernel/commit/3aa241584a26b993ef8791a28f593ca10beecdf7
https://github.com/sonyxperiadev/kernel/commit/3aa241584a26b993ef8791a28f593ca10beecdf7

humberos:
oh, ok!
Let me check charger now
Thank you

jerpelea:
https://github.com/sonyxperiadev/device-sony-common/commit/680d91ce13f15206d8e455c34e24ee029db0d693

oshmoun:
after adding system and oem mountpoints to dt, both keyaki and kugo here enter offline charging properly when connected to AC charger, so it seems to find the services now without having to move them to ramdisk.
Oh but the issue where tone devices sometimes fail to boot android still happens.

humberos:
@jerpelea 
This patch solves the mentioned race condition by itself without any additional changes.
Although we have system and odm mounted early (from dt kernel && device patches) but IMHO I still thinking this patch is a clean way to do the thing.

humberos:
@oshmoun 
Oh! crap!
Thank you for reporting back!

jerpelea:
the patch is valid for different reasons :)

PR #433, opened 2018-03-13
[COMMON] SystemUI: Disable lockscreen anti-falsing classifier

kholk:
The anti falsing feature makes some devices to be basically
impossible to get reliably unlocked at lockscreen stage and
not only: it can also be seriously difficult to even reply to a
phone call.
This is mainly seen on Loire Suzu and Tone Dora.
Disable the feature to solve the issue.


jerpelea:
if needed it should be added to the affected device

PR #435, opened 2018-03-15
Remove ro.sys.sdcardfs property

luk1337:
* It's enabled by default since 8.0.0


jerpelea:
link to the commit enabling it

luk1337:
Done.

PR #446, opened 2018-03-18
overlay: Disable rounded corners

luk1337:
* None of devices that we support have
  rounded screen.


luk1337:
Can I get some opinions about this? I don't think we really need or want the default 8dp padding, it's just unnecessary and looks kinda weird on our phones.
@jerpelea @oshmoun @alviteri

oshmoun:
would sw372dp cover all devices?

luk1337:
Yes, that's where AOSP sets 8dp padding fyi.

jerpelea:
please check your screen with and without the patch

PR #449, opened 2018-03-20
Enforce RRO on framework-res

luk1337:
Overlays only for framework-res will be converted into RROs.

Other overlays can't be converted due to some known issues on app RRO.

Bug: 36231603
Change-Id: Idca30fdbbcc990fd124de16a06a112a346612a61
(cherry picked from commit db1b6058200fc5255f519cf6609fca7619d6743e)


humberos:
This patch breaks loire!

luk1337:
try this patch on aosp.

jerpelea:
loire works fine with this patch

PR #455, opened 2018-03-21
Reenable Treble OMX

luk1337:
* Since we moved to media_profiles_V1_0.xml
  Treble OMX should be working just fine.


alviteri:
+1

PR #456, opened 2018-03-21
8.1 wifi offload

alviteri:
Test rquired @luk1337 @kholk


oshmoun:
the hal would need to be moved to hardware/wifi_offload
that's where the other hals exist

jerpelea:
I agree. It shoild be in a separate git

jerpelea:
also please indicate the source

alviteri:
moved under hardware

PR #462, opened 2018-03-25
Reorder manifest.xml

luk1337:
* Now it matches lshal order.

### DIFF
--- a/manifest.xml
+++ b/manifest.xml
@@ -103,20 +103,20 @@
         </interface>
     </hal>
     <hal format="hidl">
-        <name>android.hardware.graphics.mapper</name>
-        <transport arch="32+64">passthrough</transport>
-        <version>2.0</version>
-        <interface>
-            <name>IMapper</name>
-            <instance>default</instance>
-        </interface>
-    </hal>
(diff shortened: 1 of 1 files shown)


alviteri:
Sensors was removed due to broken sensor.qcom for Loire ping @oshmoun

luk1337:
Uh, this defines it as passthrough so no need to ping him.

alviteri:
Shouldn't be better use hwbinder ?

luk1337:
Maybe I should have called it 'Regenerate manifest.xml' instead?

alviteri:
Is ok as is called ...

oshmoun:
The sensors service can probably be added back without issues, as the repeated restarting is now fixed with https://github.com/sonyxperiadev/device-sony-common/commit/7e242d54284ea84b3cc3083e3ce9d39c5abb22fd#diff-7757ae3d5fee7079ec3e1c9d061f4d29

luk1337:
Then if you want to set it to hwbinder, do it in separate change please.

luk1337:
^^ Ideally after this change gets merged.

PR #464, opened 2018-03-25
Update lights to binder HAL

BuzzBumbleBee:


BuzzBumbleBee:
Fixed the above formatting & cast errors. Can move to more C+ file read / write in future (of if anyone wants to add it now feel free)

Tested on tone & yoshino (again needs more testing here)

PR #465, opened 2018-03-26
Implement RGB sync

luk1337:


jerpelea:
we should use the same blink for all devices

luk1337:
Yeah, someone has to do tone and yoshino then we can merge this.

jerpelea:
yep

jerpelea:
you can modify the dts for loire, tone and yoshino

jerpelea:
don't forget the LED drivers

luk1337:
Uh, I think loire isn't using qpnp led, am I right?

jerpelea:
the hal should be the same for all

PR #467, opened 2018-03-28
vold: Use mke2fs to create ext4 images

alviteri:
Set TARGET_USES_MKE2FS := true to use new EXT4 tool chain

Bug: 62421233
Test: walleye boots with new parition
Change-Id: Id6e5c68b2d7854e5330ad87443b974361a4bff9d


jerpelea:
tested devices?

PR #470, opened 2018-04-06
Copy vsync phase offsets from marlin

alviteri:
2ms/6ms is a better starting value than 1ms/1ms.

Bug: 62039896
Test: boots
Change-Id: I189d34d4204854980044546aa5c45498b979e2f9
(cherry picked from commit 14467ef10678c0f325961cea34339a989d27e645)


kholk:
These offsets are valid for 60FPS, but will make us unable to reach 120FPS in a decent manner.
We should use something like 1ms/3ms for 120FPS and we cannot change that dynamically while running the system, hence it's best to leave them as they currently are.

PR #472, opened 2018-04-09
init: remove video symlinks

alviteri:
newer devices use another video instance like video2


jerpelea:
those links were legacy

alviteri:
Were used on 8974 Hals .... Infact are not needed anymore

PR #479, opened 2018-05-08
Return "0" as duty pcts when brightness is equal 0

luk1337:
* For some reason our driver doesn't really like
  duty pcts that are like: "0,0,0,0,0,0,0,0" breaking
  custom colors in RGB sync blink.
* Returning "0" instead fixes that issue.


luk1337:
@Myself5 plzzzzzzzzz try it.

uglavnom-bezopasni:
@luk1337 @Myself5 working now :)

PR #481, opened 2018-05-12
common: Unify ro.odm.* definition

Myself5:
If we use different definitions chances are that we run into missmatches even if we don't want to.
Also, SOMC_KERNEL_VERSION is undefined.


Myself5:
As per PM, do NOT merge yet.

jerpelea:
tested

PR #482, opened 2018-05-17
hardware: qcom: bt: enable suport for sdm845

alviteri:
configure the gits for 4.9 kernel and sdm845 SOC


jerpelea:
for the initial bringup we use the msm8998 BT git

PR #486, opened 2018-05-22
netmgr: remove legacy /data/misc/netmgr folder

alviteri:
all the things were move to /data/vendor/netmgr and we should not add permissions
for the old location


jerpelea:
this has been left for backward compatibility

alviteri:
this is not used .. check logs and you will see

alviteri:
all is placed under data/vendor/netmgr

jerpelea:
use blobs v5

alviteri:
No offence but that seems stupid to use old blobs when there is v12 out ...

PR #490, opened 2018-07-07
[RQB-POWER] Sustained performance and Cluster Frequency Limits (rqb-cfl)

kholk:
Since version 7.0, Android supports a sustained performance
power hint for long-running applications, like games,
camera (in our case, viewfinder, photos but video encoding
keeps its own hint), RenderScript and audio processing:
in these cases the device requires to keep an high performance
state which will produce a lot of heat and, after a while,
the performance will decrease due to thermal mitigations.

To address this issue, handle the sustained performance hint
to limit the performance of the device to an acceptable level
while avoiding overheat.

This patchset also implements userspace management for the new
RQBalance kerneldriver's Cluster Frequency Limits functionality.

Now we support setting limits on minimum and/or maximum frequency
for all clusters by specifying that on any RQBalance-PowerHAL
profile through the XML with the parameters:
- clusterN_freq_min to limit the minimum frequency
- clusterN_freq_max to limit the maximum frequency
where N=0 stands for the "first" cluster, typically LITTLE.

Even though this is not enforced, please note that this feature
is meant to be used with a voting mechanism for temporarily
limiting one or more clusters' frequency and this shall never
be used as a way to permanently set a minimum or maximum
frequency for the clusters: for this, please refer to
the standard CPUFreq sysfs nodes.

Requires update to the RQBalance kernel driver:
https://github.com/kholk/kernel/commit/bb405414e1255377d5814c07bf8fdbb75db0eae5

---------------------------------------------------------------------------
**Example node for a Tone device**

`    <sustained_perf>`
`        <cpuquiet min_cpus="3" max_cpus="3"/>`
`        <rqbalance balance_level="40"/>`
`        <rqbalance down_thresholds="0 35 95 160 4294967295 4294967295 4294967295 4294967295"/>`
`        <rqbalance   up_thresholds="65 145 300 4294967295 4294967295 4294967295 4294967295 4294967295"/>`
`        <rqbalance    cluster0_freq_min="0"/>`
`        <rqbalance    cluster0_freq_max="1324800"/>`
`        <rqbalance    cluster1_freq_min="0"/>`
`        <rqbalance    cluster1_freq_max="1824000"/>`
`    </sustained_perf>`

---------------------------------------------------------------------------
Even though this should be *perfectly* working, this shall *not be merged until further notice*, due to undergoing testing.

This PR is here for everyone to be able to join the testing and to announce the new feature.

### DIFF
--- a/hardware/power/expatparser.c
+++ b/hardware/power/expatparser.c
@@ -17,6 +17,7 @@
 #include <errno.h>
 #include <string.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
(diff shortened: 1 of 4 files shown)


kholk:
PR forcepushed and updated to fix a bug in the XML parsing for RQB-CFL and to declare that we support the sustained performance power hint.

PR #491, opened 2018-07-09
PowerHal: move all needed perm to common

alviteri:
the same driver same paths are used by all platforms move it to common repo


alviteri:
@kholk as requested by you

kholk:
Thanks!

PR #500, opened 2018-08-03
Revert "cnss-daemon.rc: set net_admin for cnss-daemon"

alviteri:
This reverts commit 983c110205df215938629696bd2b6dc27897e2f0.


alviteri:
@jerpelea this one breaks wifi and don't let cnss.rc run

PR #502, opened 2018-08-12
[COMMON] common: init: Move hwservicemanager to avoid firmware load issues

kholk:
Move hwservicemanager start to post-fs to avoid firmware
loading issues (ADSP and SLPI) due to Android FS checks
at boot.


kholk:
Issue not solved. Closing.

PR #509, opened 2018-08-30
add guard for ro.telephony.default_network

alviteri:
at the beginning this project supported only single sim targets now there are dual sim targets
add a guard to ro.telephony.default_network so dsds devices can set own properties


jerpelea:
the propery should be overrided. please check on a dsds target that the propery is prezent in the .prop file

PR #510, opened 2018-08-31
Fix dev-keys configuration

jamuir:
PRODUCT_DEFAULT_DEV_CERTIFICATE should be used to specify the new
location of testkey.  Prior to this change, some apks were still being
signed with keys from build/make/target/product/security.

Note that setting PRODUCT_DEFAULT_DEV_CERTIFICATE does not effect what
verity key is used (so the build is still using
build/make/target/product/security/verity).

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -17,8 +17,14 @@ $(call inherit-product-if-exists, device/sony/customization/customization.mk)
 # Vendor version
 TARGET_VENDOR_VERSION := v15
 
-# Release key
-PRODUCT_DEFAULT_DEV_CERTIFICATE := vendor/oss/release-keys/releasekey
+# Specify a "dev-keys" configuration.  Keys from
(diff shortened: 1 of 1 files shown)


jamuir:
you can validate this change by doing `$ rm vendor/oss/release-keys/platform*` and then confirming that the build fails (this demonstrates that the build is indeed configured to use the various keys under `vendor/oss/release-keys`).

jamuir:
this change also has the benefit of making the post build-resigning commands work again:

```bash
$ make dist
$ ./build/tools/releasetools/sign_target_files_apks \
    -o \    # explained in the next section
    -d ~/.android-certs out/dist/*-target_files-*.zip \
    signed-target_files.zip
```

jamuir:
needs to be fixed on master, too.

jerpelea:
please PR

PR #517, opened 2018-09-11
Remove bt-hal from pathmap

luk1337:
* We are no longer building libbt-vendor
  therefore it's not necessary to set it.


alviteri:
+1

PR #520, opened 2018-09-17
8.1 updates

alviteri:


alviteri:
device tested : poplar
android-8.1.0_r40 tag

PR #522, opened 2018-09-24
sony: common: Start sensors service later

humberos:
Sensor is intermittent on suzu@loire due to some
race condition at services initialization.

So this patch is moving sensor service from main class to late_start
and also adding a property condition for its initialization.

on property:sys.qcom.devup=1

Now service starts with no problem.

Change-Id: I92aa47d073a2038c5ac71fafbe8093ba4fee0e42


humberos:
ping @oshmoun 
ping @alviteri

PR #531, opened 2018-10-22
Re-add legacy bluetooth property

ix5:
Needed for legacy devices.
See https://github.com/sonyxperiadev/bug_tracker/issues/218


ix5:
Closing since it's already fixed by https://github.com/sonyxperiadev/device-sony-common/commit/6021bcec3af0567019fe769f915a01f456578f29

PR #532, opened 2018-10-23
property: add back old BT address property

oshmoun:
Devices that have broadcom bluetooth need the old property.
Since it has no effect on newer devices with qcom bluetooth, avoid adding a guard and simply include for all devices.

### DIFF
--- a/common-prop.mk
+++ b/common-prop.mk
@@ -118,6 +118,9 @@ PRODUCT_PROPERTY_OVERRIDES += \
 # BT address
 PRODUCT_PROPERTY_OVERRIDES += \
     ro.vendor.bt.bdaddr_path=/data/vendor/bluetooth/bluetooth_bdaddr
+# BT address for devices with BCM BT
+PRODUCT_PROPERTY_OVERRIDES += \
+    ro.bt.bdaddr_path=/data/vendor/bluetooth/bluetooth_bdaddr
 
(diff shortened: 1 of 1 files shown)


oshmoun:
that can't work in a makefile afaik, it only works at runtime
but I'll test just to make sure

jerpelea:
it will be populated at runtime

oshmoun:
as suspected, it does not work:
```
kugo:/ $ getprop ro.bt.bdaddr_path                                      
$(ro.vendor.bt.bdaddr_path)
```
the value is stored as is in build.prop, and does not get resolved when called with getprop

jerpelea:
i am shure that it must expand

oshmoun:
it doesn't
actually, build just finished and I checked $OUT/system/build.prop
here's what is in there:
```
ro.vendor.bt.bdaddr_path=/data/vendor/bluetooth/bluetooth_bdaddr
ro.bt.bdaddr_path=ro.nfc.port=I2C
```
it just does not work the way you assume it would

jerpelea:
${ro.vendor.bt.bdaddr_path} and it will work

oshmoun:
if it got picked up in makefile with $() then that should also be the case with ${}
testing now anyway to know for sure

jerpelea:
please add both expanded values

PR #535, opened 2018-11-01
allow override HALs

alviteri:
needed by external project placing sony hals under hardware/*/*


alviteri:
@kholk to avoid this problem https://github.com/SonyAosp/platform_manifest/commit/6288e76459ee44fa25a4cdfee2980b3c2806e752


this PR should be enabled since i will move the things to hardware/sony

kholk:
Uhm, that looks messy.... like... a lot.

alviteri:
This let external project set own hals this will not affect the mainline project

alviteri:
Google way is messy putting hals wherever they want

PR #541, opened 2018-11-21
sony: common: Add basic USB HAL

humberos:
Add basic USB HAL that reports no status change

It depends on hardware/interfaces patch and sepolicy filecontexts entry.

Change-Id: If36c56fa3e628425cb0dc719a7d77f6e974054d1


ix5:
@humberos can you elaborate on what interfaces patch you are talking about?

And what is this fixing? I only see errors about a missing USB Gadget HAL at bootup.

ix5:
This can be closed because there is https://github.com/sonyxperiadev/vendor-oss-interfaces/ now

PR #544, opened 2018-12-06
Add QTI DIAG and RFS AIDs to config.fs

alviteri:
Bug: 79367535
Change-Id: Ic74e3663e159166bfa82a01d718aebf7cdab5f0c


jerpelea:
I could not find any trace of those AIDs 
where did you use them ?

PR #549, opened 2018-12-15
[WIP] init.common: Use /dsp instead of /vendor/dsp

ix5:
The mountpoint for the dsp partition has moved from `/dsp` to `/vendor/dsp`.

`/dsp` will remain a symlink until the migration to `/vendor` is complete.
    
Also note that calling `restorecon` on `/dsp` (without trailing slash) will result in a temporarily unlabled direcory until later in the init process, causing adsprpcd to fail.


jerpelea:
persist and dsp must be moved in each platform because we have legacy and treble devices

jerpelea:
the commit is valid for treble devices

PR #551, opened 2018-12-21
[COMMON] Keymaster 4 support

kholk:
Add support for Keymaster 4, required for KM4 devices (currently
only SoMC Tama Akatsuki).

!!!!BEWARE!!!!
This PR will break current blobs, because of the new ones having
the "vendor." prefix to the sys.listeners.registered property!
There is a wait_for_prop in the rc that have been changed to reflect
the new property!


jerpelea:
please rebase

PR #552, opened 2018-12-28
Move vendor symlink creation to modules

MarijnS95:
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/253, should fix https://github.com/sonyxperiadev/bug_tracker/issues/283 as well.

Symlink creation is moved to it's own module to make sure these are executed during the build, not during module discovery. The amount of code has been reduced with loops and variables at the same time.

In the future we might be able to declare a build rule for every library (preferrably not a `foreach` that emits a module for all) to let the build system selectively regenerate the links.

Test:
1. `m installclean` followed by `m vendorimage` on current HEAD
2. Save all vendor links to a file: `find out/target/product/akatsuki/vendor -type l -exec echo -n {}" -> " \; -exec readlink {} \; | sort -u`
3. Check out this PR and run the commands from 1. again.
4. Generate another file with the new links (command from 2.).
5. Run a diff and see that the two files (and thus all the links) are identical.

### DIFF
--- a/hardware/camera/Android.mk
+++ b/hardware/camera/Android.mk
@@ -1,21 +1,32 @@
+# Create symlinks for camera libraries. Currently only applicable
+# to sdm845 boards.
+
+LOCAL_PATH := $(call my-dir)
+
 ifneq ($(filter sdm845,$(TARGET_BOARD_PLATFORM)),)
 
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_ODM)/lib)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_ODM)/lib/camera)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_ODM)/lib64)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_ODM)/lib64/camera)
+include $(SONY_CLEAR_VARS)
+LOCAL_MODULE := camera_symlinks
+LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)
 
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/hw/camera.qcom.so hw/camera.sdm845.so && popd > /dev/null)
+library_names := \
+    camera \
+    hw/camera.qcom.so \
+    hw/com.qti.chi.override.so \
+    libcamxfdalgov7.so \
+    libcamxfdengine.so \
+    libcamxstatscore.so \
+    libcamxtintlessalgo.so \
+    libcom.qti.chinodeutils.so
 
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/hw/camera.qcom.so hw/camera.qcom.so && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/hw/com.qti.chi.override.so hw/com.qti.chi.override.so && popd > /dev/null)
+# Create symlinks to 32-bit camera libraries:
+SONY_SYMLINKS := $(foreach p,$(library_names), \
+    /odm/lib/$p:$(TARGET_COPY_OUT_VENDOR)/lib/$p \
+)
 
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/camera camera && popd > /dev/null)
+# Special exception for camera.qcom.so that is also linked to as camera.sdm845.so:
+SONY_SYMLINKS += /odm/lib/hw/camera.qcom.so:$(TARGET_COPY_OUT_VENDOR)/lib/hw/camera.sdm845.so
 
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/libcom.qti.chinodeutils.so libcom.qti.chinodeutils.so && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/libcamxfdalgov7.so libcamxfdalgov7.so && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/libcamxfdengine.so libcamxfdengine.so && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/libcamxstatscore.so libcamxstatscore.so && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/lib > /dev/null && ln -sf /odm/lib/libcamxtintlessalgo.so libcamxtintlessalgo.so && popd > /dev/null)
+include $(SONY_BUILD_SYMLINKS)
 
 endif
--- a/hardware/qca_cld3/Android.mk
+++ b/hardware/qca_cld3/Android.mk
@@ -1,4 +1,13 @@
+LOCAL_PATH := $(call my-dir)
+
 ifeq ($(WIFI_DRIVER_BUILT),qca_cld3)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/firmware/wlan/qca_cld)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /data/vendor/wifi/wlan_mac.bin firmware/wlan/qca_cld/wlan_mac.bin && popd > /dev/null)
+
+include $(SONY_CLEAR_VARS)
+LOCAL_MODULE := qca_cld3_symlinks
+LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)
+
+SONY_SYMLINKS := /data/vendor/wifi/wlan_mac.bin:$(TARGET_COPY_OUT_VENDOR)/firmware/wlan/qca_cld/wlan_mac.bin
+
+include $(SONY_BUILD_SYMLINKS)
+
 endif
--- a/hardware/tftp/Android.mk
+++ b/hardware/tftp/Android.mk
@@ -1,36 +1,43 @@
+LOCAL_PATH := $(call my-dir)
+
 ifneq ($(filter sdm660 msm8998 sdm845,$(TARGET_BOARD_PLATFORM)),)
 
-$(shell rm -r $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/rfs/)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/rfs/msm/mpss/readonly/vendor)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/rfs/msm/adsp/readonly/vendor)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/rfs/msm/slpi/readonly/vendor)
-$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/rfs/msm/cdsp/readonly/vendor)
+include $(SONY_CLEAR_VARS)
+LOCAL_MODULE := tftp_symlinks
+LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)
 
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /data/vendor/tombstones/modem rfs/msm/mpss/ramdumps && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /persist/rfs/msm/mpss rfs/msm/mpss/readwrite && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /persist/rfs/shared rfs/msm/mpss/shared && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /persist/hlos_rfs/shared rfs/msm/mpss/hlos && popd > /dev/null)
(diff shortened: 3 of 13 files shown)


jamuir:
looks good, but I think there are other symlink shell commands inside device-sony-common that need to be similarly changed.

MartinX3:
Thank you for your work :)
Then I will test it after every symlink shell commands got "upgraded" to avoid any missing symlink which would lead to a bug/crash or something. :)

jamuir:
I like the idea of creating something analogous to `BOARD_ROOT_EXTRA_SYMLINKS`, but I guess that would require a patch to `platform/build`.  The sony tree doesn't currently patch that project, but it could be done with an update to repo_update.sh.

Limiting the changes to inside device-sony-common seems less complicated.

jamuir:
@jerpelea :  is a patch to `platform/build` something that you might consider?

jamuir:
@MarijnS95 :  I didn't realize that the implementation of `BOARD_ROOT_EXTRA_SYMLINKS` was inside `system/core`; I see now that it is done using `LOCAL_POST_INSTALL_CMD`.  There would be no need to patch `platform/build` if you wanted to add something analogous for symlinks inside the vendor dir.  Sorry for the distraction.

MarijnS95:
@jamuir The third commit adds a little build module abstraction within the Sony common tree to generate symlinks. Looking forward to opinions as I'm not satisfied with the extra overhead of this method.

On the other hand, a global variable that's picked up at the end might hard to ~do~ get right. If ordering changes (accidentally), symlinks are not created when they are appended to such a global variable after the pickup script has ran.

Similarly, I don't think we can make a module that expands a variable at build time (and thus picks up everything).

sony_build_symlinks.mk
```diff
@@ -0,0 +1,16 @@
+# Creates symlinks relative to the PRODUCT_OUT folder.
+# This is different from BOARD_ROOT_EXTRA_SYMLINKS in that it allows symlinks on other partitions.
+#
+# Links can be specified in TARGET:DIRECTORY or TARGET:LINK_NAME notation. This creates a link named LINK_NAME or DIRECTORY/$(basename TARGET) pointing to TARGET.
+
+ifeq ($(SONY_SYMLINKS),)
+    $(warning "No symlinks set!")
+else
+LOCAL_POST_INSTALL_CMD := $(foreach s,$(SONY_SYMLINKS), \
+    $(eval p := $(subst :,$(space),$(s))) \
```
ix5: Why not use a space here in the first place?

jamuir: well, using `$(space)` is easier to read.

the code here is taken from `system/core/rootdir/Android.mk`, and they use `$(space)`.

jamuir:
This looks very good!

Since it follows the familiar aosp build-module pattern, it should be easy for others to use.

I don't think we really need to worry about the overhead of re-creating the symllnks every time "make" runs.

sony_build_symlinks.mk
```diff
@@ -0,0 +1,16 @@
+# Creates symlinks relative to the PRODUCT_OUT folder.
+# This is different from BOARD_ROOT_EXTRA_SYMLINKS in that it allows symlinks on other partitions.
+#
+# Links can be specified in TARGET:DIRECTORY or TARGET:LINK_NAME notation. This creates a link named LINK_NAME or DIRECTORY/$(basename TARGET) pointing to TARGET.
+
+ifeq ($(SONY_SYMLINKS),)
+    $(warning "No symlinks set!")
+else
```
jamuir: Maybe add a comment here pointing readers to `system/core/rootdir/Android.mk` in case they want to compare.

sony_build_symlinks.mk
jamuir: re-reading your comment, I guess you meant why not use inputs of the form `t1 n1  t2 n2 t3 n3 ...` instead of `t1:n1  t2:n2 t3:n3 ...`

ix5: Yup, unless whitespace has some special meaning here, it seems more sensible to just do that directly.
In the context of `BOARD_ROOT_EXTRA_SYMLINKS` Google decided to use the `SOURCE:TARGET` syntax and then convert that to `SOURCE TARGET` here, but since we are contructing the file list automatically, we can just skip that step.

If however we want to add `SOURCE:TARGET` pairings manually later, it would make sense to keep Google's convention.

MarijnS95: You are both missing the fact that the initial foreach loops over SONY_SYMLINKS with multiple TARGET:LINK_NAME pairs separated by spaces. If the `TARGET` and `LINK_NAME` are also separated by spaces, you need to add a lot of extra trickery to combine every second element with the first. This is why `:` is used to group the two together.

jamuir: I didn't miss that fact :-)

In any case, my preference is to leave it as it is.  The colon-separator is a familiar pattern in aosp makefiles.

jamuir:
that is surprising that setting `LOCAL_VENDOR_MODULE := true` doesn't cause the sym-link modules to be picked up by `m vendorimage`.  Maybe switching to a different build module type might help (i.e.  using something other than `BUILD_PHONY_PACKAGE`).

hardware/camera/Android.mk
```diff
@@ -1,21 +1,31 @@
+# Create symlinks for camera libraries. Currently only applicable
+# to sdm845 boards.
+
+# Warning: The same guard is used in common.mk, where this module
+# is depended on.
 ifneq ($(filter sdm845,$(TARGET_BOARD_PLATFORM)),)
```
jamuir: why not just drop this guard?

hardware/qca_cld3/Android.mk
```diff
@@ -1,4 +1,9 @@
 ifeq ($(WIFI_DRIVER_BUILT),qca_cld3)
```
jamuir: same here.

hardware/tftp/Android.mk
```diff
@@ -1,36 +1,29 @@
 ifneq ($(filter sdm660 msm8998 sdm845,$(TARGET_BOARD_PLATFORM)),)
```
jamuir: and here.

hardware/tftp/Android.mk
```diff
@@ -1,36 +1,29 @@
...
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /persist/hlos_rfs/shared rfs/msm/slpi/hlos && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /firmware rfs/msm/slpi/readonly/firmware && popd > /dev/null)
-$(shell pushd $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR) > /dev/null && ln -sf /odm/firmware rfs/msm/slpi/readonly/vendor/firmware && popd > /dev/null)
+# Prepend vendor and prefix directory to all link names:
+SONY_SYMLINKS := $(foreach prefix,$(TARGET_PREFIXES), \
+    $(foreach s,$(TARGET_COMBINATIONS), \
+        $(eval p := $(subst :,$(space),$(s))) \
+        $(word 1,$(p)):$(TARGET_COPY_OUT_VENDOR)/$(prefix)/$(word 2,$(p)) \
+    ) \
+)
```
jamuir: this is fairly complex.

writing a verbose list of the form `t1:n1 t2:n2 ...` here might be better since it easier to maintain.

The advantage of just keeping a flat list of the form  is that it is very

sony_build_symlinks.mk
```diff
@@ -0,0 +1,16 @@
...
+LOCAL_POST_INSTALL_CMD := $(foreach s,$(SONY_SYMLINKS), \
+    $(eval p := $(subst :,$(space),$(s))) \
+    mkdir -p $(dir $(PRODUCT_OUT)/$(word 2,$(p))); \
+    ln -sf $(word 1,$(p)) $(PRODUCT_OUT)/$(word 2,$(p)); \
+)
+endif
```
jamuir: I wrote this comment earlier (and deleted it), but here it is again:  it might be helpful to include a comment pointing to `system/core/rootdir/Android.mk` so people know where to find the original implementation.

hardware/camera/Android.mk
MarijnS95: These guards had to be dropped from `common.mk` instead, because these variables (`TARGET_BUILD_PLATFORM` and `WIFI_DRIVER_BUILT`) are set *after* `common.mk` has been fully included.
Luckily, adding modules to `PRODUCT_PACKAGES` does not result in errors when these do not exist (which is weird considering this can easily lead to typos and other accidents).

hardware/tftp/Android.mk
MarijnS95: As explained in the commit description, this is done intentionally. From what it looks these are all the same, and converting it to an outer product enforces that. Reducing typos at the same time.

However, you may be right in that I had to add an edgecase for the tombstones which do not follow this pattern (accidentally?). And the case does not hold for the `readwrite` symlinks. Again, this may be intentional, but it could easily be the typo this code is preventing against:

(Diff comparing current `HEAD` to the new symlink creation, generated with `find out/target/product/akatsuki/vendor -type l -exec echo -n {}" -> " \; -exec readlink {} \; | sort -u`)
```diff --git a/old_symlinks b/new2_symlinks
index c2f3085..d53c970 100644
--- a/old_symlinks
+++ b/new2_symlinks
@@ -205,13 +205,13 @@ out/target/product/akatsuki/vendor/rfs/msm/adsp/hlos -> /persist/hlos_rfs/shared
 out/target/product/akatsuki/vendor/rfs/msm/adsp/ramdumps -> /data/vendor/tombstones/lpass
 out/target/product/akatsuki/vendor/rfs/msm/adsp/readonly/firmware -> /firmware
 out/target/product/akatsuki/vendor/rfs/msm/adsp/readonly/vendor/firmware -> /odm/firmware
-out/target/product/akatsuki/vendor/rfs/msm/adsp/readwrite -> /persist/rfs/msm/adsp
+out/target/product/akatsuki/vendor/rfs/msm/adsp/readwrite -> /persist/rfs/msm/mpss
 out/target/product/akatsuki/vendor/rfs/msm/adsp/shared -> /persist/rfs/shared
 out/target/product/akatsuki/vendor/rfs/msm/cdsp/hlos -> /persist/hlos_rfs/shared
 out/target/product/akatsuki/vendor/rfs/msm/cdsp/ramdumps -> /data/vendor/tombstones/rfs/cdsp
 out/target/product/akatsuki/vendor/rfs/msm/cdsp/readonly/firmware -> /firmware
 out/target/product/akatsuki/vendor/rfs/msm/cdsp/readonly/vendor/firmware -> /odm/firmware
-out/target/product/akatsuki/vendor/rfs/msm/cdsp/readwrite -> /persist/rfs/msm/cdsp
+out/target/product/akatsuki/vendor/rfs/msm/cdsp/readwrite -> /persist/rfs/msm/mpss
 out/target/product/akatsuki/vendor/rfs/msm/cdsp/shared -> /persist/rfs/shared
 out/target/product/akatsuki/vendor/rfs/msm/mpss/hlos -> /persist/hlos_rfs/shared
 out/target/product/akatsuki/vendor/rfs/msm/mpss/ramdumps -> /data/vendor/tombstones/modem
@@ -223,5 +223,5 @@ out/target/product/akatsuki/vendor/rfs/msm/slpi/hlos -> /persist/hlos_rfs/shared
 out/target/product/akatsuki/vendor/rfs/msm/slpi/ramdumps -> /data/vendor/tombstones/rfs/slpi
 out/target/product/akatsuki/vendor/rfs/msm/slpi/readonly/firmware -> /firmware
 out/target/product/akatsuki/vendor/rfs/msm/slpi/readonly/vendor/firmware -> /odm/firmware
-out/target/product/akatsuki/vendor/rfs/msm/slpi/readwrite -> /persist/rfs/msm/slpi
+out/target/product/akatsuki/vendor/rfs/msm/slpi/readwrite -> /persist/rfs/msm/mpss
 out/target/product/akatsuki/vendor/rfs/msm/slpi/shared -> /persist/rfs/shared```

@jerpelea Is this correct, or should all `*/readwrite` links point to their own `/persist/rfs/msm/*` folder?

File: sony_build_symlinks.mk (hunk 7 above)
Comment (MarijnS95) [reply sony_build_symlinks.mk]: Forgot about it, comment is now added up-top.

**Comment by MarijnS95** [comment]
This pull request is now done from my perspective. All review comments have been handled and regressions have been tested again and addressed (sorry for the typo mess with force-pushing and editing the title thrice).

However, we still have to substitute other `shell` commands in `hardware/firmware/Android.mk` and `misc/version/Android.mk` (the latter being only of interest for @jerpelea). Should I address these in this PR?

**Comment by MartinX3** [comment]
@MarijnS95 
So this bug not only occur creating the adreno symlinks?
Missing symlinks will occur at other places, too?

If yes, can you @MartinX3 me in the other PR's or in this PR again. after it is done and can be tested?

**Comment by MarijnS95** [comment]
@MartinX3 You can test it right now. That makefile only creates 3 folders:
```makefile
$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/dsp)
$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/firmware_mnt)
$(shell mkdir -p $(PRODUCT_OUT)/$(TARGET_COPY_OUT_VENDOR)/bt_firmware)
```
You can create these manually, or `touch hardware/firmware/Android.mk` to regenerate and have them created for you.

MartinX3:
@MarijnS95
Thank you for this information.
One question I have is: how to cherrypick this pullrequest easily?
I only know how to cherrypick a gerrit change request.

jerpelea:
please do them all

jerpelea:
please address them all

hardware/camera/Android.mk
ix5: I think the build system only complains about missing modules if `LOCAL_MODULE_TAGS := optional` is not set?

ix5:
> @MarijnS95 
> One question I have is: how to cherrypick this pullrequest easily?
> I only know how to cherrypick a gerrit change request.

https://help.github.com/articles/checking-out-pull-requests-locally/

MarijnS95:
@ix5 Ninja'd it :grin: 

I usually use `git fetch sony pull/<id>/head && git merge FETCH_HEAD`, saves loose branches lying around.
(That command is assuming you have `sony` as a remote configured to this repository. If not, substitute it with the right remote name or the clone url of the repo).

MartinX3:
Thank you @ix5 
"command line instructions" is not available at my site, because I am not a repository contributor.
So I managed to apply this pull request as a patch. :)

MarijnS95:
@MartinX3 You do not need to be a contributor or have access to *fetch* or *pull* commits. That's only for *pushing* new changes.

MartinX3:
@MarijnS95 
Yes, sorry, English is not my native language and sometimes I use the wrong words. :)

MarijnS95:
@MartinX3 Oh I wasn't pointing at your English (which is fine btw), just the slight nuances GitHub uses to determine different "types" of users.
Feel free to hover over that tag on Alin's and my comments; it'll tell you the difference right away :)

Anyway, this PR comment section is getting a little too chatty. With the last push all `shell` commands have been replaced with modules. @jerpelea Please test the change to odm version generation.

However, the phony modules have another annoying side-effect. `LOCAL_MODULE_PATH` has to be set to make sure the build system recognizes that it's part of the vendor image, but that causes the `<LOCAL_MODULE>-timestamp` file to be placed there: ugly and confusing.

MartinX3:
@MarijnS95
@jerpelea
Nice, are there still places with missing symlinks or can I start building the rkom again? :)

jamuir:
@MarijnS95: would you like me to open a PR that backports this to the oreo branch?

jerpelea:
please do

MarijnS95:
@jamuir Already on it, there are a few conflicts to resolve.

MarijnS95:
@jamuir See https://github.com/sonyxperiadev/device-sony-common/pull/555. Can you test it and compare the original symlinks with the new ones just in case? There are a few minor differences (with Pie).

@MartinX3 Yes you are safe to rebuild; there are no other `shell` commands modifying the state in the output directories.

PR #554, opened 2019-01-06
[WIP] Add new health HAL and remove healthd

ix5:
<strike>The new health HAL lives at https://github.com/ix5/vendor-sony-oss-health for now.</strike>

This PR also removes the old board-specific `libhealthd` library, which was largely unused on Pie anyway.

## Remove board libhealthd library

The [custom patches to system/core/healthd to enable `healthd_mode_charger_*` functions](https://android-review.googlesource.com/c/platform/system/core/+/373812) were dropped in Pie (they're no longer present in repo_update.sh) and as such the board libhealthd library is unused apart from `healthd_board_init()` and `healthd_board_battery_update()`, which can safely be left to their AOSP defaults.

The default AOSP `libhealthd_charger` can draw charging animations and set LEDs on its own already, so we no longer need to patch it or draw via custom `healthd_board_mode_charger_*` functions.

## Remove deprecated backup healthd instance

Our new health HAL enabes use to move to health@2.0 and deprecate healthd entirely.

For now, AOSP ships a backup instance of healthd by default, so disable it in accordance with the official documentation: https://source.android.com/devices/tech/health/deprecation

## Notes
If you still see `healthd` stanzas in your logs, that's because of this:
- We depend on  the shared_lib `android.hardware.health@2.0-impl`, which wants `libbatterymonitor`, see [Android.bp](https://android.googlesource.com/platform/hardware/interfaces/+/android-9.0.0_r22/health/2.0/default/Android.bp#25) and [Health.cpp](https://android.googlesource.com/platform/hardware/interfaces/+/android-9.0.0_r22/health/2.0/default/Health.cpp#38)
- `libbatterymonitor` is [provided by BatteryMonitor.cpp](https://android.googlesource.com/platform/system/core/+/android-9.0.0_r22/healthd/Android.bp#10)
- [BatteryMonitor.cpp](https://android.googlesource.com/platform/system/core/+/android-9.0.0_r22/healthd/BatteryMonitor.cpp#17) logs as `healthd`

Once AOSP's `android.hardware.health@2.0-impl` gets rid of `libbatterymonitor`, which they've stated they intend to, we'll get rid of it too automatically.


jerpelea:
the new healthd should be pushed into the common git

ix5:
Moving to common as requested

PR #558, opened 2019-01-20
common-treble: Drop unnecessary wifi, fingerprint and radio interface inclusions.

MarijnS95:
These are just interface definitions which are used to compile various
modules. The generated helper libs contain boilerplate code to convert
from function calls to binder parcels and back. These are already
depended upon explicitly by the services themselves and thus included in
the final build.
Since there's no point in depending on these again (only to add
confusion and mix versions), drop all the lines.

EDIT: This is a continuation of #557, with more libraries dropped and a different commit description. Thanks @ix5 for pointing me to these dangling dependencies.

### DIFF
--- a/common-treble.mk
+++ b/common-treble.mk
@@ -20,10 +20,9 @@ PRODUCT_PACKAGES += \
     android.hardware.configstore@1.1-service
 
 # RIL
+# Interface library needed by odm blobs:
 PRODUCT_PACKAGES += \
-    android.hardware.radio@1.2-radio-service \
-    android.hardware.radio.config@1.0 \
-    android.hardware.radio.deprecated@1.0
+    android.hardware.radio.config@1.0
 
 # Audio
 PRODUCT_PACKAGES += \
@@ -39,8 +38,6 @@ PRODUCT_PACKAGES += \
 
 # Wi-Fi
 PRODUCT_PACKAGES += \
-    android.hardware.wifi@1.0 \
-    android.hardware.wifi@1.1 \
     android.hardware.wifi@1.0-service
 
 # NFC packages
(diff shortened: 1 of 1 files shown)


ix5:
Same as #563, I think we have to keep `radio.deprecated` for now.

MarijnS95:
@ix5 Do you know by chance which platform uses it? I can't find any references to the deprecated library in the Loire/Nile/Tama v5 odm image, and our build tree only references it in `hardware/ril/libril`. That latter one is only built when `BOARD_PROVIDES_LIBRIL` is set, which is not the case for any of our platforms/devices.

That said, the removal runs fine on all devices of the aforementioned platforms.

ix5:
It's the OemHook stuff in TransPower and some other places afaik.

MarijnS95:
First, disregard the previous comment. It's `ifneq ($(BOARD_PROVIDES_LIBRIL),true)`, meaning that libril is definitely built in the tree. Lucky us, we can drop `android.hardware.radio.deprecated@1.0` since that library explicitly depends on it.

Looking at this further, here are some observations:

- `IOemHook` is registered in `harware/ril/libril/ril_service.cpp`.
- The addition of the default `radio.config@1.0-service` in #563 doesn't make much sense. There are references to `IRadioConfig` and it's `::registerAsService` in `libril-qc-qmi-1.so`, but do not seem to be called. Which complements the fact that it's not registered (as seen in `service list`) either. The functions in the default service are no-ops anyway.
  In any case, we are not specifying this service to be available in our manifest.xml.
- Same for `android.hardware.radio@1.2-sap-service `: `ISap` is registered in `hardware/ril` (depending on the available sims even), so there's no point to add this no-op service (the functions are stubs).
- We are specifying the wrong radio versions. `hardware/ril/libril` registers `IRadio` and `ISap` 1.1, but we declare 1.2.
- `service list` (and/or `vndservice list`) are broken. The aforementioned services are definitely successfully registered (added log statements to print the return status), but do not show up in the list.

Which means this PR is good as it is. I'd still like to continue on #563, but do the following differently:
- Change our manifest to reflect `IRadio` and `ISap` 1.1.
- Leave out the no-op services.
- Check if `libril-qc-qmi-1.so` is really registering `IRadioConfig`. If so, we should specify it in the manifest for correctness.
- Keep an `fqname` for `IOemHook` as it's still specified and used.

jerpelea:
radio service should be in use

PR #559, opened 2019-01-20
Adapt LED flashing time to android-dictated pause-times

MarijnS95:
Android provides on- and off-times in `LightState` when requesting LED flash (for notifications). Because our default timings (set in dt) are rather short and appear trippy, adhere to the framework-orchestrated timings which are more forgiving 

TODO:
- Create and merge PR's to allow access to these new files. I'll get them up soon.
- This assumes all rgbsync (those with `/sys/class/leds/rgb/rgb_blink`) have these pause properties available. Please confirm that before merging.
  I have not seen this property on Suzu, which doesn't have rgbsync either. Those platforms should stay unharmed.
- Test *without* the last commit. Scaling rgb channels with the `brightness` attribute is more efficient than scaling and applying the entire `duty_pcts` curve, but doesn't seem to work on akatsuki at least (this could also be a bug in the qpnp led driver).

Tested on SoMC Akatsuki DSDS.

### DIFF
--- a/hardware/liblights/Light.cpp
+++ b/hardware/liblights/Light.cpp
@@ -271,6 +271,15 @@ namespace android {
                                 writeStr(RED_LED_DUTY_PCTS_FILE, getScaledDutyPcts(red));
                                 writeStr(GREEN_LED_DUTY_PCTS_FILE, getScaledDutyPcts(green));
                                 writeStr(BLUE_LED_DUTY_PCTS_FILE, getScaledDutyPcts(blue));
+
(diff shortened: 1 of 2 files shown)


MarijnS95:
@jerpelea This PR is ready from my point of view. All ownership changes have been merged, and the brightness idea is removed because it does not work (checked the driver, it uses only the pwm strength from `duty_pcts` and never blends it with `brightness`, so we'll have to keep updating those for now).

PR #560, opened 2019-01-20
allow override default_network common prop on dsds

alviteri:
dsds targets have own ro.telephony.default_network that one from common is not needed to copy in build.prop


oshmoun:
you could simply use ?= instead
that should suffice to avoid an overwrite

PR #564, opened 2019-01-26
[DONOTMERGE] adreno symlinks: Update vulkan lib name

ix5:
For vendor v5, the file was renamed from `vulkan.$platform.so` to `vulkan.qcom.so`.

### DIFF
--- a/hardware/adreno/Android.mk
+++ b/hardware/adreno/Android.mk
@@ -19,13 +19,18 @@ library_names := \
     libllvm-qcom.so \
     librs_adreno.so \
     librs_adreno_sha1.so \
-    hw/vulkan.$(TARGET_BOARD_PLATFORM).so
+    hw/vulkan.qcom.so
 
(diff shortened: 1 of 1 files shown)


hardware/adreno/Android.mk
```diff
@@ -28,4 +28,9 @@ SONY_SYMLINKS := $(foreach lib_dir,lib lib64, \
     ) \
 )
 
+# Special exception for vulkan.qcom.so that is also linked as vulkan.$(TARGET_BOARD_PLATFORM).so
+SONY_SYMLINKS := $(foreach lib_dir,lib lib64, \
```
MarijnS95: `:=` overwrites the previous value of `SONY_SYMLINKS`, meaning that all other links defined above are lost. Please change it to `+=`.

MarijnS95: Also, thanks @erikcas for noticing the missing backslash on the next line.

PR #556, opened 2019-01-07
manifest: Add tetheroffload config and control hals.

MarijnS95:
Hal definitions from https://android.googlesource.com/device/google/crosshatch/+/618a01fe100bd98902e59ee9447cae975d4ce3b4

We still have to adapt the HAL itself to build for properly for the respective platforms.

PR #568, opened 2019-01-29
[COMMON] [VERIFY_ME] init.common.rc: Create directory for IPACM

kholk:
Creating it from the ipacm.rc doesn't appear to work for
some reason, hence create it here along with the other
directories.

At least on SoMC Loire Suzu, that directory didn't get
created for some reason.
I've had to create it manually to stop ipacm from constantly
restarting and failing due to that missing dir.

It's a very strange issue, so please verify it before merging.

### DIFF
--- a/rootdir/vendor/etc/init/hw/init.common.rc
+++ b/rootdir/vendor/etc/init/hw/init.common.rc
@@ -179,6 +179,9 @@ on post-fs-data
 
     # RIL directoy for services
     mkdir /data/vendor/radio 0771 system radio
+    
+    # Create directory for IPACM
+    mkdir /data/vendor/ipa 0771 radio radio
 
     chown system /dev/block/bootdevice/by-name
 


MarijnS95:
@kholk This is expected. Currently the directory is created in `on boot`, which is too late. See `system/core/rootdir/init.rc` for the full details. But for example, the `main` service class (which `ipacm` is part of) is started from `on zygote-start` when the phone is unencrypted. `zygote-start` runs **after** `post-fs-data` but before `boot`.
https://github.com/sonyxperiadev/device-sony-common/pull/556/commits/33bb1f6b3359b642dc92ac9cc5de0c92d8433c2a Solves the issue by moving it to `post-fs-data`, which is also what all the other scripts use. Let me update the commit description to cover this case.

kholk:
Hehehe yes putting it there was on purpose.
I will check tomorrow, maybe I haven't got your commit up locally.

I've posted this pr as "verify me" because this was one of the possible solutions to a problem that seemed very very odd to me, so that can be more like a reminder for me and not only... :P

jerpelea:
fixed by https://github.com/sonyxperiadev/device-sony-common/commit/0922397bf1fa7e01cb606176959c633822995882

PR #569, opened 2019-01-31
[DONOTMERGE] Adjust permission xml files target location

ix5:
- Use `TARGET_COPY_OUT_SYSTEM` instead of "system" so system-as-root targets won't have issues later on

- Copy vendor app perms to vendor: ExtendedSettings and TransPower are vendor apps now, so copy the accompanying priv-app permission xml files to `TARGET_COPY_OUT_VENDOR`

### DIFF
--- a/common-perm.mk
+++ b/common-perm.mk
@@ -56,8 +56,8 @@ PRODUCT_COPY_FILES += \
 
 # Transmit power
 PRODUCT_COPY_FILES += \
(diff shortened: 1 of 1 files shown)


ix5:
Don't merge this until https://github.com/sonyxperiadev/transpower/pull/7 and https://github.com/sonyxperiadev/packages_apps_ExtendedSettings/pull/42 are merged or we will have boot failures!

PR #571, opened 2019-02-02
config: configure brightness level

Gray47Maxx:
This commit will drastically improve battery life while device is being used with a minimum brightness level. Tested with suzu, hopefully should work on all platforms.


ix5:
This is the wrong repo to PR this change. Look at the individual device repos, e.g. https://github.com/sonyxperiadev/device-sony-kagura/blob/master/overlay/frameworks/base/core/res/res/values/config.xml#L70

Gray47Maxx:
Hm, I thought changing this will override all other platforms' settings...

But thanks anyways, I think I'll redo these changes for all platforms then today if this is 100% inappropriate commit

PR #572, opened 2019-02-06
common overlay: Add UMTS to radio access family

ix5:
Just out of curiosity, is there a reason UMTS is not on that list?


jerpelea:
deprecated

PR #575, opened 2019-02-16
[COMMON] Binderized RQBalance-PowerHAL

kholk:
This is a restructured RQBalance-PowerHAL for the new
binderized structure found in (currently) Android 9.0 Pie.

It currently supports the same hints as the previous version
of this HAL, but wires up (planned) support for new hints
as found in the IPower 1.1, 1.2, 1.3.
The new hints are currently blank.

Tested on SoMC Tama Akatsuki RoW.

P.S.: This needs changing all the devices .mk's to build this new HAL,
which has _obviously_ got a different name, compared to the old one!

### DIFF
--- a/hardware/power/Android.mk
+++ b/hardware/power/Android.mk
@@ -48,9 +48,6 @@ LOCAL_SHARED_LIBRARIES := \
     libhardware \
     libhidlbase \
     libhidltransport \
-    android.hardware.power@1.0 \
-    android.hardware.power@1.1 \
-    android.hardware.power@1.2 \
     android.hardware.power@1.3
 
 include $(BUILD_EXECUTABLE)
--- a/hardware/power/Android.mk
+++ b/hardware/power/Android.mk
@@ -24,12 +24,26 @@ LOCAL_HEADER_LIBRARIES := libhardware_headers
 
 LOCAL_C_INCLUDES := external/expat/lib
 
-LOCAL_SRC_FILES := power.c rqbalance_halext.c expatparser.c
-LOCAL_SHARED_LIBRARIES := liblog libcutils libexpat
-LOCAL_MODULE := power.$(TARGET_DEVICE)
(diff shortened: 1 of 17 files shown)


hardware/power/android.hardware.power@1.3-service.sony.rc
```diff
@@ -0,0 +1,5 @@
+service vendor.power-hal-1-3 /vendor/bin/hw/android.hardware.power@1.3-service.sony
+    class hal
```
ix5: Add `interface android.hardware.power@1.3::IPower default` here, might be useful in the future, not only for [dynamic hals](https://source.android.com/devices/architecture/hal/dynamic-lifecycle#changes-HAL-definitions)

kholk: That's a nice one, thanks, I will :D

hardware/power/Android.mk
```diff
@@ -24,12 +24,26 @@ LOCAL_HEADER_LIBRARIES := libhardware_headers
...
+    liblog \
+    libcutils \
+    libexpat \
+    libdl \
+    libbase \
+    libutils \
+    libhardware \
+    libhidlbase \
+    libhidltransport \
+    android.hardware.power@1.0 \
```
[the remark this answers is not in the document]
ix5: Seems I was wrong, the old lib definitions are still needed

kholk:
So, I need to either stub out properly or to really read the low power stats, then this PR will be ready to merge.

Currently, with proper stubs, on Akatsuki, the HAL is working perfectly fine.

kholk:
The HAL now supports both stubs and real low power stats and the whole thing should be _**ready for prime time**_.

Please test it on at least one legacy platform before merging.

Enjoy!

infixremix:
Don't we need this for vibrator Hal as well?

PR #576, opened 2019-02-22
health: Do not spam logcat with battery stats

ix5:
The setting was left enabled in the testing phase but can be turned off now.


jerpelea:
the spm is usefull

PR #577, opened 2019-02-23
Drop android.hardware.power@1.0 service and impl

luk1337:
* Since we are providing our own service and
  power.$(TARGET_DEVICE) is dead there's no
  reason to build it.

### DIFF
--- a/common-packages.mk
+++ b/common-packages.mk
@@ -99,8 +99,7 @@ PRODUCT_PACKAGES += \
 
 # OSS Power HAL
 PRODUCT_PACKAGES += \
-    librqbalance \
-    android.hardware.power@1.3-service.sony
+    librqbalance
 
 # OSS WIFI and BT MAC tool
 PRODUCT_PACKAGES += \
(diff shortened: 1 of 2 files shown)


MarijnS95:
Before this gets merged, notice that the 1.3 power HAL was [added to `common-packages.mk`](https://github.com/sonyxperiadev/device-sony-common/commit/7e1074ee4ca150c434197cb92329c6f98b94c9ac) instead of the usual `common-treble.mk`. Do we want to correct this for consistency?

luk1337:
Umm
https://github.com/sonyxperiadev/device-sony-common/commit/7e1074ee4ca150c434197cb92329c6f98b94c9ac

luk1337:
I don't want to correct this for consistency. I'm just removing it cause only a single `android.hardware.power/IPower` interface can be registered and supposedly it's not always that the power HAL we want registers it first. At least according to slack messages in #general.

kholk:
Solves issues _at least_ when Android wants to use power@1.2 interfaces.

This PR is CRITICAL. MERGE!

PR #579, opened 2019-03-03
init: Update vendor services

ix5:
- Add `interface` information to services' `init.rc` files, allows for dynamic HALs and maybe more in Q
- Rename services to follow the `vendor.*` pattern as upstream does
- Remove unused `charger.rc` definition

Most of these changes bring our configs in line with crosshatch.

**Attention:** Test carefully whether some blob relies on the old naming scheme, e.g. for `qseecomd`.


ix5:
Tested, should be fine to merge now.

PR #581, opened 2019-03-03
Enforce usage of vintf manifest

ix5:
Check manifest and compatibility_matrix files at build time

### DIFF
--- a/CommonConfig.mk
+++ b/CommonConfig.mk
@@ -93,6 +93,10 @@ TARGET_NO_RPC := true
 BOARD_CHARGER_DISABLE_INIT_BLANK := true
 BOARD_CHARGER_ENABLE_SUSPEND := true
 
+# Check manifest and compatibility_matrix files at build time
+PRODUCT_ENFORCE_VINTF_MANIFEST_OVERRIDE := true
+VINTF_ENFORCE_NO_UNUSED_HALS := true
+
 # Enable dex-preoptimization to speed up first boot sequence
 ifeq ($(HOST_OS),linux)
   ifneq ($(TARGET_BUILD_VARIANT),eng)


ix5:
Running this for about 2 weeks without issues, may need testing for tama with keymaster stuff?

ix5:
The side effect of this change is that existing "DualSIM Patchers" will no longer work, since the second instance of the HALs is not in the single-SIM manifest.
The patchers will have to include updated `manifest.xml` files (but don't worry, we're on this already).

ix5:
Ready for merge.

MarijnS95:
Running this since the creation of the PR on devices from pretty much all platforms (Suzu, Discovery, Mermaid, Akatsuki), no regressions. Has the go-ahead from me :+1:

Paulbouchara:
Running fine on my devices (kagura, maple, Akatsuki)
Good to go !

ix5:
It's kind of funny, because current `master`/Q throws an error for `radio@1.2` even without this build flag set... ;)

But yeah, this buildvar is enabled on all devices with `SHIPPING_API_LEVEL >= 26` anyway (see [build/make/core/config.mk](https://android.googlesource.com/platform/build/+/refs/tags/android-q-preview-6/core/config.mk#672)), so @MarijnS95 I'd be very surprised if you saw any difference :)

This PR is mainly for legacy devices, and I've tested extensively on tone.

MarijnS95:
Which basically means enforcement is "useless" on Pie, as we've seen before. Good to know that Q reminds us though ^^

ix5:
Already done by #640 
`VINTF_ENFORCE_NO_UNUSED_HALS` is an internal build system var and should not be set.

PR #582, opened 2019-03-03
vintf: Add vendor.nxp.INxpNfc HAL to manifest

ix5:
`INxpNfc` is a vendor extension that is supported by the HAL code in `hardware/nxp/*`.


jerpelea:
please add description

ix5:
Updated to remove `INxpEse` since there is no secure element on SODP(see #584)

PR #584, opened 2019-03-09
[WIP] common-treble: Add NXP Secure Element HAL

ix5:
`android.hardware.secure_element` is used for "secure" credential transfer for e.g. chip cards.


jerpelea:
we dont have a secure element!

PR #587, opened 2019-03-13
common-packages: Build SimDetect

luk1337:
* This package will show a dialog asking user
  to reboot their device whenever they remove
  or insert a SIM card.


luk1337:
Not yet (I guess?); It should be added to local_manifests and sepolicy repo first I suppose.

PR #588, opened 2019-03-19
[WIP] common-prop: Prefix "vendor." to rild.libpath

ix5:
This is needed when switching to whitelisted "compatible" properties only.


jerpelea:
remove the old one

ix5:
Done, but should only be merged when the transition is complete.

jerpelea:
only after v7

PR #593, opened 2019-03-20
common-prop: Add back rild.libpath

ix5:
The legacy prop is still needed for devices which cannot be made fully
"compatible"

### DIFF
--- a/common-prop.mk
+++ b/common-prop.mk
@@ -134,6 +134,7 @@ PRODUCT_PROPERTY_OVERRIDES += \
 
 # RILD
 PRODUCT_PROPERTY_OVERRIDES += \
+    rild.libpath=/odm/lib64/libril-qc-qmi-1.so \
     vendor.rild.libpath=/odm/lib64/libril-qc-qmi-1.so \
     ril.subscription.types=NV,RUIM
 


jerpelea:
please add ifdef for COMPATIBLE

ix5:
You cannot make an ifdef for product variables.
Any variable in the [_product_var_list](https://android.googlesource.com/platform/build/+/refs/tags/android-9.0.0_r34/core/product.mk#75) is only available in later stages as the `device.mk` file is evaluated lazily.
Both `PRODUCT_SHIPPING_API_LEVEL` and `PRODUCT_COMPATIBLE_PROPERTY` are on the list so they cannot be evaluated in `common-prop.mk` as they are empty at that stage, even if you define `PRODUCT_SHIPPING_API_LEVEL` in `device.mk` (or rather, they get "internalized" when inheriting products and the product inherited has no access to the buildvar).

The workaround would be to define a custom variable like `TARGET_LEGACY_NONCOMPATIBLE` or whatever, but that's stupid since there is no harm in keeping the legacy prop.

When we have fully switched all devices to compatible props we can get rid of this PR again, but we have to keep the legacy prop for now.

PR #596, opened 2019-04-01
common: allow skipping setting PRODUCT_DEFAULT_DEV_CERTIFICATE

oshmoun:
Some projects want to use their own keys, so make it possible to skip defaulting to a specific key.


jerpelea:
there is no need for this 
I moved the certificate to a separate git

PR #597, opened 2019-04-20
common-prop: avoid double encryption

lazerl0rd:
https://source.android.com/devices/storage/adoptable#fixing-double-encryption


lazerl0rd:
Alongside https://github.com/sonyxperiadev/kernel/pull/1962.

PR #598, opened 2019-04-21
init: Boot DSP before SLPI again

ix5:
init: Boot DSP before SLPI again

Before 8792c97, adsp and cdsp were booted before slpi:

    init.qcom.devstart.sh
    echo 1 > /sys/kernel/boot_adsp/boot
    echo 1 > /sys/kernel/boot_cdsp/boot
    echo 1 > /sys/kernel/boot_slpi/boot

8792c97 changed that behaviour to first booting SLPI via `init.qcom.devstart.sh`, setting `vendor.qcom.devup=1` and only then booting ADSP/CSDP via `init.qcom.(adsp|cdsp)start.sh`.

That and later commits caused race conditions that meant audio was gone on tone devices every other boot.

This commit restores the old behaviour of first booting ADSP/CDSP and then SLPI.

### DIFF
--- a/rootdir/vendor/etc/init/adspstart.rc
+++ b/rootdir/vendor/etc/init/adspstart.rc
@@ -4,6 +4,3 @@ service adspstart_sh /vendor/bin/init.qcom.adspstart.sh
     group root system
     disabled
     oneshot
-
-on property:vendor.qcom.slpiup=1
-    start adspstart_sh
--- a/rootdir/vendor/etc/init/cdsprpcd.rc
+++ b/rootdir/vendor/etc/init/cdsprpcd.rc
@@ -6,4 +6,3 @@ service vendor.cdsprpcd /odm/bin/cdsprpcd
 
 on property:vendor.qcom.cdspup=1
     enable vendor.cdsprpcd
-
--- a/rootdir/vendor/etc/init/cdspstart.rc
+++ b/rootdir/vendor/etc/init/cdspstart.rc
@@ -4,6 +4,3 @@ service cdspstart_sh /vendor/bin/init.qcom.cdspstart.sh
     group root system
     disabled
     oneshot
-
-on property:vendor.qcom.slpiup=1
-    start cdspstart_sh
--- a/rootdir/vendor/etc/init/hw/init.common.rc
+++ b/rootdir/vendor/etc/init/hw/init.common.rc
@@ -88,8 +88,11 @@ on fs
 
     mount_all /vendor/etc/fstab.${ro.hardware}
 
-    # Start devices as soon as partitions are moounted
-    start slpistart_sh
+    # Start DSP services as soon as partitions are mounted
+    # ADSP devices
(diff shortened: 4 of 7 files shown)


ix5:
Updated PR with the following changes:

`rootdir/vendor/bin/init.qcom.adspstart.sh` and `cdspstart.sh`:
Add 2s sleep delay after booting before setting `vendor.qcom.adspup` (or `cdspup`)

`rootdir/vendor/etc/init/adsprpcd.rc`:
Changed `class late_start` to `class main` for `vendor.adsprpcd`
Start the service immediately, don't just enable it:
```
 on property:vendor.qcom.devup=1
+    start vendor.adsprpcd
     enable vendor.adsprpcd
```

`rootdir/vendor/etc/init/sensors.rc`:
Start `vendor.sensors` `on property:vendor.qcom.devup=1` and not on `adspup`. This means sensors will only be started after SLPI is up.

ix5:
Tested successfully on kagura, but needs testing on other platforms.

ix5:
Updated again to remove the `start vendor.adsprpcd` line, no need since `class_start main` is triggered before `devup` is set.

ix5:
With further results rolling in it appears that this PR isn't needed after all.
Still leaving open for visibility since we should start the services in the correct order.

ix5:
Updated to remove the 2s sleep delay. Tested for weeks on kagura with no regressions.

rootdir/vendor/etc/init/devstart.rc
```diff
@@ -4,3 +4,11 @@ service devstart_sh /vendor/bin/init.qcom.devstart.sh
     group root system
     disabled
     oneshot
+
+# ADSP devices
+on property:vendor.qcom.adspup=1
```
MarijnS95: Shouldn't this be condensed into `on property:vendor.qcom.adspup=1 && property:vendor.qcom.cdspup=1`? That way both devices are guaranteed to be started before starting SLPI.

On devices that don't have `cdsp` this doesn't matter, because their script will (even when failing to `echo 1` to the file) set `cdspup` to `1`.

ix5: The `init.cdspstart.sh` and `cdsprcpd.rc` files only get copied for sdm845 devices
See https://github.com/sonyxperiadev/device-sony-common/blob/master/rootdir/Android.mk#L262-L279

MarijnS95: Disregard the last sentence. This'll be a race condition only on sdm845 then.

MarijnS95: Consensus: Let's not worry about this. Tama is already fine with `SLPI` started _before_ adsp/cdsp, so it shouldn't matter if it's started inbetween those.

ix5:
Tested exhaustively, also on tama. Fixes urgent adsp race on tone.
MERGEME

PR #599, opened 2019-04-26
[WIP] sensors.rc: Add restorecon for /persist/sensors/

ix5:
Fix labels when coming from stock

Note: The `restorecon` in `hw/init.common.rc` may skip the persist folder since it only checks whether the sepolicy `file_contexts` file has changed and not the actual folder contents.


ix5:
(See also https://github.com/sonyxperiadev/device-sony-common/pull/600 as an alternative)

oshmoun:
confirmed working on an affected device

MartinX3:
Sorry, yes I confirm.

PR #600, opened 2019-04-27
[WIP] Add manual restorecon script for /persist

ix5:
The `restorecon_recursive` directive in init is only applied if the `file_contexts` file changed between builds, but not necessarily if any file or folder inside `/mnt/vendor/persist/` has changed.

The restorecon code checks whether an xattr named `security.restorecon_last` contains a string that matches the current combined hashes of the SELinux context files and skips restoring labels
if there is a match, see https://android.googlesource.com/platform/external/selinux/+/refs/tags/android-9.0.0_r35/libselinux/src/android/android_platform.c#1546

Force wiping that xattr so that restorecon always runs since it's not very expensive (there are currently only about 50 files on /persist).

The restorecon is needed to fix issues such as wrong stock labels on `/mnt/vendor/persist/sensors/`:
`sensors_persist_file` -> `persist_sensors_file`

ATTENTION: restorecon_persist.sh needs accompanying SELinux labels and contexts!
-> See https://github.com/sonyxperiadev/device-sony-sepolicy/pull/521

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -84,6 +84,10 @@ PRODUCT_PACKAGES += \
     rdclean.sh
 endif
 
+# Partition fixes
+PRODUCT_PACKAGES += \
+    restorecon_persist.sh
+
 # Qcom scripts
 PRODUCT_PACKAGES += \
     init.qcom.devstart.sh \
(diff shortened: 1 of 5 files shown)


ix5:
(This is an alternative to https://github.com/sonyxperiadev/device-sony-common/pull/599 )

PR #603, opened 2019-05-12
sensors.rc: Explicitly chown /persist/sensors

ix5:
The builtin init `chown` is not recursive, so individual filenames need to be supplied.

Fixes: `sensors.qcom` and `android.frameworks.sensorservice` cannot start up due to files in `/persist/sensors` being owned by `root:root`.


ix5:
Tested on kagura, MERGEME please

PR #605, opened 2019-05-16
[WIP] common-prop: Enable direct boot from charger mode

ix5:
Enable regular booting when long-pressing power button in charger mode
See [system/core/healthd/healthd_mode_charger.cpp](https://android.googlesource.com/platform/system/core/+/refs/tags/android-9.0.0_r39/healthd/healthd_mode_charger.cpp#431)


ix5:
broken at least on kagura, nvm

PR #608, opened 2019-05-20
[DONOTMERGE] [Q-COMPAT] Remove gps_debug.conf

ix5:
**DO NOT MERGE** Not tested on P yet.

frameworks/base/services/core/Android.bp already copies frameworks/base/services/core/java/com/android/server/location/gps_debug.conf to /system/etc/gps_debug.conf.
The file is identical to ours and the duplication throws a build error on Android Q.


ix5:
Already done by #640

PR #610, opened 2019-05-22
[Q-COMPAT] Move kernel includes to common.mk

ix5:
But keep the include for the prebuilt kernel in `CommonConfig.mk` since its makefile depends on `$(TARGET_DEVICE)` which is only set after `common.mk` has been read.

`common.mk`: Make `BUILD_KERNEL` customizable
Use `?=` instead of `:=` so it is possible to override the buildvar in `customization.mk`.


ix5:
Already done via #640

PR #615, opened 2019-05-31
[Q-COMPAT] Power: Add interface info to .rc

ix5:
Android's servicemanager somehow still wants to start @1.0 at boot, so annotate that version as well.
Tested on Q/master, please confirm on P.

(Also sneak in the removal of the unused `st` var in `Hints.cpp`)


ix5:
Ready for merge.

PR #617, opened 2019-05-31
ODM creation refactoring

ix5:
Use `PRODUCT_ODM_PROPERTIES` for version - avoids invoking shell from make.

Note: The oem partition's `build.prop` has moved to `/odm/etc/build.prop` on the master branch / on Android Q, but `ro.odm.version` will still be read.

(This one I'm willing to debate, @jerpelea your call)
Only build if SONY_BUILD_ODM is set
This should be set in `customization.mk` if building the blobs for `odm.img`.

Avoids having non-existent `PRODUCT_PACKAGES` when doing OSS AOSP builds.


jerpelea:
I agree with the clean way 
ifdef it !

ix5:
You mean `ifeq ($(SONY_BUILD_ODM),true)` ?

ix5:
Also moved the keymaster+gatekeeper products to `common-odm`.

Also, I'm wondering why `keymaster@4` doesn't include an `-impl`, I guess it is it statically linked?
https://github.com/sonyxperiadev/device-sony-common/blob/master/common-treble.mk#L92-L102

ix5:
Need feedback from you @jerpelea whether this works for you.

ix5:
Ping @jerpelea

common-odm.mk
```diff
@@ -113,6 +113,17 @@ PRODUCT_PACKAGES += \
     libqsocket \
     qmuxd
 
+ifneq ($(TARGET_LEGACY_KEYMASTER),true)
+ifneq ($(TARGET_KEYMASTER_V4),true)
+# Keymaster
+PRODUCT_PACKAGES += \
+    android.hardware.keymaster@3.0-impl-qti
+endif
```
jerpelea: please move all keymaster qti blobs !

ix5: The `-service-qti` "products" aren't actually blobs, they're "virtual" packages as defined in [`rootdir/Android.mk`](https://github.com/sonyxperiadev/device-sony-common/blob/fd3e6c8c993d3aa7ef7ae9856d37dc09d4bbcf3f/rootdir/Android.mk#L27), so they do not belong in the odm image creation file.

We need those "virtual" packages in _every_ build since they supply the `-service-qti.rc` files in `/vendor/etc/init/`.

MarijnS95:
> Also, I'm wondering why `keymaster@4` doesn't include an `-impl`, I guess it is it statically linked?

The `-impl` is a same-process library that can be used directly through `dlopen` + `dlysym("HIDL_FETCH_IKeymaster..")`, or `defaultPassthroughServiceImplementation` for binderized mode. Keymaster is most likely not used as same-process HAL (only binderized) so I guess they dropped the extra lib in favor of implementing directly in `-service-qti`

However, this proves a fair question: What do we need `-service-qti` for? If there is a same-process implementation, we can just bundle the default `-service` from AOSP that loads the same-process library to provide functionality.
Any reason for not moving these to odm btw?

MarijnS95:
Tested the above claim and it works perfectly fine on Mermaid (KM4) and Discovery (KM3). The internals for `keymaster@3.0-service-qti` and `gatekeeper@1.0-service-qti` are _identical_ to AOSP.

This means we can remove two useless "proprietary" blobs and replace them with something OSS (despite being quite useless to look at :no_mouth:). More importantly, we can clean up a lot of cruft:

- Prune sepolicy
  - (I do not really understand why we have files for `hal_{keymaster,gatekeeper}_default` already, since this HAL is not even used by us)
  - It is quite curious that the HALs operate fine (encryption and pincode work) while running under a little more constrained sepolicy; there is no access to firmware files... After all the separate executable `qseecomd` is responsible for opening and pushing these files to the TZ.
- Delete the files from the `odm` builds
- Delete rc files
- Delete `-service-qti` "prebuilt" targets (except km4)

But we have to be a little careful:
- `interface` definitions in `.rc` only exist in master, not on the Android 10 tags (not that it matters, it is not a dynamic HAL)
- `keymaster` has `user nobody`, and is not in the `system` group
- `gatekeeper` is started as hal, not `early_hal`

jerpelea:
this would be awesome but has to be tested and patches cherrypicked

MarijnS95:
Deploying on the Daily Driver as we speak. Will get a thorough test.
@ix5 Can you prepare the patches and update this one, as `[WIP]`?

MarijnS95:
For the adventurous: you can test this by cherry-picking these commits:
https://github.com/MarijnS95/device-sony-common/commit/50e319f3b3bdae28fcfcb83a5a28b7d7fbf8c64a
https://github.com/MarijnS95/device-sony-sepolicy/commit/2139fd1bac5d9c74c94acf731a279f015b52808c
https://github.com/MarijnS95/device-sony-sepolicy/commit/8604b4dc3eebfd28515c2c039e106d62519a7424

Note I have done something very ugly. KM4 runs in its own QTI service, but works just fine with the `hal_keymaster_default` policy (after all, that's what the passthrough implementations will be using now as well).
The odm images still label this executable as `hal_

This has been tested on Akatsuki (KM4), Mermaid (KM4) and Discovery (KM3), all encrypted and with pincodes. Still needs a check on a legacy keymaster device, where a `KM[0-2]` legacy libhardware implementation is loaded and exposed as `keymaster@3.0`.

PR #619, opened 2019-06-03
Remove non-existent packages

ix5:
Remove `p2p_supplicant.conf`: This package does not exist.

Remove `libemoji`: This package does not exist since Pie.
See https://android.googlesource.com/platform/frameworks/opt/emoji/+refs

common-treble: nfc: Remove `@1.1-impl`: It does not exist, not even in Q.
Why not include `@1.0-impl`?
`@1.1-service` already includes the implementation, so no need to include an unused older `-impl`.

Remove `imsservices.rc`: No recent device sets `TARGET_HAS_IMS` any more.


ix5:
Ready for merge.

PR #620, opened 2019-06-03
[WIP] common-packages: Add netutils-wrapper

ix5:
netutils-wrapper is a wrapper for netd to call /system tools such as /system/bin/ip from a vendor context.
This avoids selinux denials from netd and similar tools.

(Needs testing on Pie)

### DIFF
--- a/common-packages.mk
+++ b/common-packages.mk
@@ -79,6 +79,10 @@ PRODUCT_PACKAGES += \
     wpa_supplicant.conf \
     libwpa_client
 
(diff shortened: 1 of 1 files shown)


oshmoun:
the wrapper is included here without having to explicitly add it as a package

ix5:
You're right, see [build/make/target/product/full_base.mk](https://android.googlesource.com/platform/build/+/refs/tags/android-9.0.0_r40/target/product/full_base.mk#42)

PR #621, opened 2019-06-16
[2/3] Add fingerprint navigation config files

MarijnS95:
For https://github.com/sonyxperiadev/vendor-sony-oss-fingerprint/pull/50

These config files are used to tie uinput events coming from `uinput-fpc` to semantic names that are understood by Android.


overlay/frameworks/base/core/res/res/values/config.xml
```diff
@@ -261,4 +261,7 @@
 
     <!-- Flag indicating whether we should enable smart battery. -->
     <bool name="config_smart_battery_available">true</bool>
+
+    <!-- True if the device supports system navigation keys. -->
+    <bool name="config_supportSystemNavigationKeys">true</bool>
```
ix5: This should be moved to platform overlays since we cannot support loire and tone.

MarijnS95: That is entirely correct, and a remnant of hoping we could support all of them.

Who knows if we find a workaround before the thing is even tested and merged on other platforms :wink:

PR #622, opened 2019-07-03
common.mk: Force split props for all devices

ix5:
Move all vendor props into `(/system)/vendor/build.prop`
Mainly for treble compliance but also to declutter the platform repos that already define this.

This buildvar can be set for all devices, it just doesn't change that much for older devices without a phsyical `/vendor` partition.


ix5:
Preparing full treble for all devices on Q.
Ready for merge.

PR #624, opened 2019-07-03
Remove unused irsc_util and sec_config

ix5:


ix5:
Unless `sec_config` is used somewhere deep down in the blobs, I think this can go.
@jerpelea unless you have objections, this is ready for merge.

jerpelea:
it is used!

PR #625, opened 2019-07-03
common-prop: Remove unused ro.nfc.port

ix5:


ix5:
Can't find any reference to this. Seems to be a remnant from the Gingerbread days.
Ready for merge.

PR #635, opened 2019-08-30
[DONOTMERGE] [Q-ONLY] common: Put more RROs into /vendor

ix5:
Only merge into Q branch!

---

Since between the various devices we are modifying more overlays than just `framework-res`, we want our modifications to be treble-compatible and land in the `/vendor` partition as `$APPNAME__auto_generated_rro.apk`.

Since https://r.android.com/810894 it is possible to add other targets than `framework-res` to the list of the packages for which RROs are generated.

This commit aids treble compliance so that our devices retain their overlays when running GSIs.


ix5:
Already done by #640

PR #636, opened 2019-08-31
sony-common: Enable 64 bit DRM plugins compilation

alviteri:
This will generate DRM plugins, HALs & oemcrypto_test in 64 bit.

Change-Id: I451db39edea3446d9816fcaba77caa8d3b097afb


alviteri:
Test done with Netflix and mermaid .. working

PR #609, opened 2019-05-20
[DONOTMERGE] [Q-COMPAT] qcom/utils.mk: Use ?= for definitions

ix5:
**DO NOT MERGE** Not tested on P yet.

"empty", "space" etc. might already be defined and a redefinition throws a build error on Android Q.

PR #640, opened 2019-09-10
[Android 10] Welcome Android Q!

MarijnS95:
Android Q is finally here and SODP is one of the first to enjoy the new features such as _system wide darkmode_ and the _mess_ that is the new gesture mode! (Do you want to open an app drawer? Hah, you just clicked back :grin:).

Contains cherry-picks from #608, #609, #610 #635

This patchset cleans up redefinitions, removes 

#### `BUILD_KERNEL` and `TARGET_COMPILE_WITH_MSM_KERNEL`
As far as I'm aware these variables mean pretty much the same thing. The former is used to compile the kernel, the latter to compile and include headers (usually?). Instead of making an attempt to merge them together, set `TARGET_COMPILE_WITH_MSM_KERNEL := $(BUILD_KERNEL)`. These will most likely always be used in conjunction.

#### Enable VNDK on all platforms
Not all platforms suffice the minimum version (sdk `28`), thus override this to true to have strict rules everywhere.

#### New bootctrl (and `soong_namespace`...)
Android now supports shared library HALs in recovery. The new bootctrl uses this, together with namespaces to separate (possibly overlapping, see Ganges and Nile) module definitions with the same name.
However, in order to use these packages in "old style" make (eg. `make <target>` from the command line or `PRODUCT_PACKAGES`) the namespace needs to be explicitly imported (which makes sense, the build system needs to know which package you want). See the relevant commit for more information.


ix5:
The `gps.conf` removal might have been premature, because we're now using the Qcom default `SUPL` servers instead of SOMC's.

MarijnS95:
@ix5 That's a good point. @jerpelea Do we want to see about overriding that one? I don't think there's a way to tell the build system to force this file through, meaning a patch to the GPS HAL just for this.

jerpelea:
copy the new gps.conf here

MarijnS95:
@jerpelea That won't work since we are getting a duplicate definition for `gps.conf`.

Not to worry, there's a nice solution:
https://android-review.googlesource.com/c/platform/hardware/qcom/gps/+/204644

jerpelea:
i meant that you should copy the file here

MarijnS95:
@jerpelea Do you want the gps.conf copied into _this conversation_ to take a look at the difference? Since there's quite a bit.

PR #646, opened 2019-09-26
common: Enforce all overlays to be a separate RRO on /vendor.

MarijnS95:
For example SystemUI didn't show up despite the name being correct.
Instead of trying to collect all currently overriden targets, and not
forgetting to add them in the future, follow crosshatch and build _any_
overlay as RRO. This is the most compatible with GSIs.


ix5:
I second this, it's much saner. There are no RROs we could want exempt.

PR #628, opened 2019-07-14
[1/2] [WIP] health: Add support for storage stats and health

ix5:
## health: Add support for storage stats and health

This implements the missing health@2.0 function calls:
- `get_storage_info()`
- `get_disk_stats()`

Also introduce a `HealthTestingMock` for debugging purposes. It is only available in userdebug/eng builds.
The prop `persist.vendor.health.fake_storage=1` will trigger fake stats.

Storage health reporting is not available for UFS since there is no kernel support for UFS health as of 2019-07-12.

Since the SODP devices all have different emmc nodes, attempt to resolve `/sys/block/mmcblk0` to e.g.
`/sys/devices/platform/soc/7464900.sdhci/mmc_host/mmc0/mmc0:0001/block/mmcblk0`, which gives use the emmc node at `/sys/devices/platform/soc/7464900.sdhci/mmc_host/mmc0/mmc0:0001`

For UFS devices, construct the sysfs paths based on the property `ro.boot.bootdevice`, which would look like e.g. `7464900.ufshci`, giving us a path such as: `/sys/devices/platform/soc/7464900.ufshci`.
(Currently unused of course)

The nodes will have to labeled in the individual platforms' sepolicy dirs:
- [common sepolicy](https://github.com/sonyxperiadev/device-sony-sepolicy/pull/530)
- [tone](https://github.com/sonyxperiadev/device-sony-tone/pull/183)
- ...

SODP devices report pre-EOL info, two lifetime values and an EMMC revision.

Disk statistics are taken from the generic `stat` file, just as the default implemention of `storaged` and the crosshatch/bonito health HAL do.

## health: Refactor and simplify, drop healthd shims

<strike>This change drastically simplifies the code, which had been written under the assumption that AOSP would drop healthd soon. The wrapper libraries make little sense, just drop them.</strike>
**Update:** The "simplify" commits have been split out into https://github.com/sonyxperiadev/device-sony-common/pull/647

PR #647, opened 2019-09-30
health: Simplify

ix5:
Split out from https://github.com/sonyxperiadev/device-sony-common/pull/628 to get the refactor merged first.

Tested on kagura but should be fine on all devices.


ix5:
Tested on Android Q as well now.

PR #649, opened 2019-10-01
vintf: Upgrade wifi hal to 1.3

sjllls:
All work fine on both 1.2 and 1.3, we update it for more new features

### DIFF
--- a/vintf/manifest.xml
+++ b/vintf/manifest.xml
@@ -239,7 +239,7 @@
     <hal format="hidl">
         <name>android.hardware.wifi</name>
         <transport>hwbinder</transport>
-        <version>1.2</version>
+        <version>1.3</version>
         <interface>
             <name>IWifi</name>
             <instance>default</instance>


MarijnS95:
Well, thanks for making me rebase again :no_mouth:

jerpelea:
always a pleasure :)

sjllls:
> Well, thanks for making me rebase again no_mouth
> 
> Also, were are updating this version first and foremost for _correctness_. The default wifi HAL implements v1.3, and our manifest should reflect that.

Yeah, should we upgrade [this](https://github.com/sonyxperiadev/device-sony-common/blob/master/common-treble.mk#L41)?
Thanks for your suggestion !

MarijnS95:
@sjllls No, I mean that _that_ (correctness) is the reason for updating to 1.3, not "just because we can" or "because maybe new features". Doesn't matter now, commit has been merged already.

jerpelea:
we should move to 1.3 for the new features that we will soon use

MarijnS95:
@jerpelea We have already moved to 1.3, with this bump in the vintf.

Google hardly ever updates the _minor_ HAL version number, I think to save on required edits elsewhere. Instead, the default `1.0` HAL implements the V1.3 HIDL.

sjllls:
> @jerpelea We have already moved to 1.3, with this bump in the vintf.
> 
> Google hardly ever updates the _minor_ HAL version number, I think to save on required edits elsewhere. Instead, the default `1.0` HAL implements the V1.3 HIDL. (There's your answer @sjllls wink)
> 
> Here I am again hoping for Google to migrate to the new `vintf_fragments` option, as this allows the manifest version to be bundled and properly incremented with the implementation itself, where it belongs.

Thanks a lot!

PR #650, opened 2019-10-05
[1/2] common: Replace gatekeeper/keymaster passthrough blobs with AOSP variant

MarijnS95:
These blobs only provide a passthrough service for the implementations
stored in the shared libraries. AOSP provides an identical solution
complete with .rc files and "OSS code" (not that there's anything to see
apart registering a default passthrough implementation).

See
https://github.com/sonyxperiadev/device-sony-common/pull/617#issuecomment-536089919
for an expanded explanation.

### DIFF
--- a/common-treble.mk
+++ b/common-treble.mk
@@ -71,29 +71,26 @@ PRODUCT_PACKAGES += \
 PRODUCT_PACKAGES += \
     android.hardware.biometrics.fingerprint@2.1-service.sony
 
+# Gatekeeper passthrough service
+PRODUCT_PACKAGES += \
+    android.hardware.gatekeeper@1.0-service
+
 ifeq ($(TARGET_LEGACY_KEYMASTER),true)
 # Keymaster
 PRODUCT_PACKAGES += \
     android.hardware.keymaster@3.0-impl \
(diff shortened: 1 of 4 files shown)


common-treble.mk
```diff
@@ -71,29 +71,26 @@ PRODUCT_PACKAGES += \
...
+    android.hardware.gatekeeper@1.0-impl
 else ifeq ($(TARGET_KEYMASTER_V4),true)
-# Keymaster 4
+# Keymaster 4 passthrough service
 PRODUCT_PACKAGES += \
-    android.hardware.keymaster@4.0-service-qti
-# Gatekeeper
-PRODUCT_PACKAGES += \
-    android.hardware.gatekeeper@1.0-service-qti
+    android.hardware.keymaster@4.0-service
```
MarijnS95: KM4 Must still be `-qti`, let me fix+text+push this.

jerpelea:
seems to work but I will not merge it until we have all devices tested with it

ix5:
Booting fine on kagura (tone), which uses keymaster v3.

jerpelea:
tested on nile

PR #652, opened 2019-10-20
common-treble/vintf: Remove widevine refs

ix5:
The widevine service is proprietary and not shipped with AOSP.
Declaring it in the `vintf` manifest will result in an endless loop of `servicemanager` trying to start the service.

Downstream projects are advised to revert this commit if they are shipping the widevine blobs.


MarijnS95:
> Downstream projects are advised to revert this commit if they are shipping the widevine blobs.

I can also recommend vintf fragments within a blueprint `cc_prebuilt_binary` :wink:

jerpelea:
do it

PR #654, opened 2019-11-12
gpt-utils: dereference symlinks pointing to device path

rinigus:
This PR fixes the issue of getting the correct device name in Sailfish. In contrast to AOSP, Sailfish populates /dev with relative and not absolute links. As a result, current `get_dev_path_from_partition_name` was failing. This implementation gets a symlink, de-references it, and then guesses device name on the basis of de-referencing result.

Let me know if this is of any interest and, if it is, I should submit it towards master branch.

Acknowledgements: Basic idea is coming from Adam Pigg implementation for one of Sailfish ports, implementation discussed with Marijn

### DIFF
--- a/hardware/gpt-utils/gpt-utils.cpp
+++ b/hardware/gpt-utils/gpt-utils.cpp
@@ -1072,12 +1072,14 @@ static int get_dev_path_from_partition_name(const char *partname,
                 if (stat(path, &st)) {
                         goto error;
                 }
-                if (readlink(path, buf, buflen) < 0)
-                {
+                char resolved[PATH_MAX] = {0};
+                if (!realpath(path, resolved)) {
                         goto error;
-                } else {
-                        buf[PATH_TRUNCATE_LOC] = '\0';
                 }
+                int pos = (int)strlen(resolved) - 1;
+                for (; pos >= 0 && isdigit(resolved[pos]); pos--) ;
+                resolved[pos + 1] = '\0';
+                strncpy(buf, resolved, buflen);
         } else {
                 snprintf(buf, buflen, BLK_DEV_FILE);
         }


jerpelea:
please submit to master

rinigus:
Thanks, closing it here

hardware/gpt-utils/gpt-utils.cpp
```diff
@@ -1072,12 +1072,14 @@ static int get_dev_path_from_partition_name(const char *partname,
...
-                if (readlink(path, buf, buflen) < 0)
-                {
+                char resolved[PATH_MAX] = {0};
+                if (!realpath(path, resolved)) {
                         goto error;
-                } else {
-                        buf[PATH_TRUNCATE_LOC] = '\0';
                 }
+                int pos = (int)strlen(resolved) - 1;
+                for (; pos >= 0 && isdigit(resolved[pos]); pos--) ;
```
MarijnS95: Since this loop has no body and no initializer, I think it's generally more desirable to write `while (pos >= 0 && isdigit(resolved[pos])) pos--;`.

hardware/gpt-utils/gpt-utils.cpp
```diff
@@ -1072,12 +1072,14 @@ static int get_dev_path_from_partition_name(const char *partname,
                 if (stat(path, &st)) {
                         goto error;
                 }
-                if (readlink(path, buf, buflen) < 0)
-                {
+                char resolved[PATH_MAX] = {0};
```
MarijnS95: Is there a particular reason to use a separate buffer on the stack, and copy back to `buf` after the fact? The original code used `buf` directly as well.
When encountering an error and returning it, any caller is hopefully supposed to treat `buf` as containing undefined data.

hardware/gpt-utils/gpt-utils.cpp
rinigus: preparing now new PR, but I'll reply here.

yes, that would make sense. I'll make this change before submitting

hardware/gpt-utils/gpt-utils.cpp
rinigus: `buf`has no guarantees regarding its length. at the same time, `realpath` expects PATH_MAX sized string. So, this is to avoid some odd out-of-memory error induced if `buf` is shorter.

So far, the calling code had `buf` length set to PATH_MAX, but I haven't checked it systematically. and since API doesn't force it, we cannot assume `buf` length either.

MarijnS95: That is true, unfortunately `realpath` expects the buffer to be at least `PATH_MAX` bytes in size. Though returning an error when it's smaller - even when `PATH_MAX` is too much to realistically fit the string - might not be desirable. For correctness it should return an error when `buflen < pos + 2` before doing the `strncopy`, though.

rinigus: Thanks, I will add it. Will probably submit it tomorrow - would like to run the corrected code on device as well

PR #655, opened 2019-11-12
gpt-utils: dereference symlinks pointing to device path

rinigus:
This PR fixes the issue of getting the correct device name in Sailfish. In contrast to AOSP, Sailfish populates /dev with relative and not absolute links. As a result, current `get_dev_path_from_partition_name` was failing. This implementation gets a symlink, de-references it, and then guesses device name on the basis of de-referencing result.

I have tested earlier version on Sailfish for some time now and it works as expected. I have not tested on AOSP, though. The last version has `for` loop replaced by `while` - haven't tested that on device yet. But its expected to be equivalent to the tested one.

Acknowledgements: Basic idea is coming from Adam Pigg implementation for one of Sailfish ports, implementation discussed with Marijn

### DIFF
--- a/hardware/gpt-utils/gpt-utils.cpp
+++ b/hardware/gpt-utils/gpt-utils.cpp
@@ -1079,6 +1079,9 @@ static int get_dev_path_from_partition_name(const char *partname,
                 int pos = (int)strlen(resolved) - 1;
                 while (pos >= 0 && isdigit(resolved[pos])) pos--;
                 resolved[pos + 1] = '\0';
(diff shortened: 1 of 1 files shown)


rinigus:
I have added the buffer length check, as suggested last night. Please review and test on AOSP. It works as expected on Sailfish

rinigus:
Thanks for merging!

jerpelea:
please pr on p-mr1

rinigus:
As an additional PR for it as well?

PR #661, opened 2019-11-28
[COMMON] Switch to QTI Display Allocator/Mapper 3.0

kholk:
This patchset switches all devices to the new QTI Allocator and Mapper 3.0.

Requires @MarijnS95's display HAL PR and https://github.com/sonyxperiadev/vendor-qcom-opensource-interfaces/pull/6


kholk:
@MarijnS95 has pointed out that my fast switching is too fast, hence too dirty and he will redo it the clean way.

Closing this PR as he'll open one himself to achieve this result!

PR #663, opened 2019-11-28
CommonConfig: Set VENDOR_SECURITY_PATCH for new Keymasters.

MarijnS95:
Google added a new property ro.vendor.build.security_patch that's not
set to any sensible value by default. Our kew keymaster implementations
(KM4 on kernel 4.14 for the time being) rely on this value for rollback
protection.

Set it to the same value as the system currently built.

@ix5 Do you have any comment on this? Seems the only sensible default to me, no idea why AOSP doesn't set it.

### DIFF
--- a/CommonConfig.mk
+++ b/CommonConfig.mk
@@ -138,3 +138,7 @@ endif
 ifeq ($(AB_OTA_UPDATER),true)
 DEVICE_MANIFEST_FILE += $(COMMON_PATH)/vintf/android.hardware.bootctrl.xml
 endif
+
+# New vendor security patch level: https://r.android.com/660840/
+# Used by newer keymaster binaries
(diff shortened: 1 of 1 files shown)


ix5:
> @ix5 Do you have any comment on this? Seems the only sensible default to me, no idea why AOSP doesn't set it.

My speculation is that it isn't being set in order for vendors to override it, but then again, they could just use `?=` in `build/make`.

FYI Google pretty much does the same thing [device/google/bonito:device-common.mk](https://android.googlesource.com/device/google/bonito/+/refs/tags/android-10.0.0_r10/device-common.mk#38).

MarijnS95:
> FYI Google pretty much does the same thing [device/google/bonito:device-common.mk](https://android.googlesource.com/device/google/bonito/+/refs/tags/android-10.0.0_r10/device-common.mk#38).

Cool, then we did the right thing. Might add that boot SPL as well, just in case.

PR #668, opened 2019-12-04
[COMMON] [URGENT] CommonConfig: Remove kernel cross compile prefixes for Q

kholk:
Android 10's build system will fail to build the kernel if we use these
prefixes due to "unallowed paths".


ix5:
Ugh you and your PE crap... :P
It works fine on regular AOSP with a reworked kernel `Android.mk`.

PR #675, opened 2019-12-20
[WIP] common: Simplify treble buildvars, add VNDK pkg

ix5:
Since all devices are now fully treble compatible, we can enable `PRODUCT_FULL_TREBLE_OVERRIDE`.

`PRODUCT_FULL_TREBLE_OVERRIDE` sets:
- `PRODUCT_TREBLE_LINKER_NAMESPACES`
- `PRODUCT_SEPOLICY_SPLIT`
- `PRODUCT_ENFORCE_VINTF_MANIFEST`
- `PRODUCT_NOTICE_SPLIT`
And also `BOARD_PROPERTY_OVERRIDES_SPLIT_ENABLED` indirectly.

Setting `PRODUCT_USE_VNDK_OVERRIDE` is still needed because `PRODUCT_USE_VNDK` will only be set based on `PRODUCT_FULL_TREBLE` if the shipping API level is 27 or greater.

Also add `vndk_package` in common so that devices don't need to add it individually.

We might get rid of these vars entirely by inheriting products from `build/make/product/` at a later date.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -33,17 +33,19 @@ PRODUCT_ENFORCE_RRO_TARGETS := *
 
 PRODUCT_DEXPREOPT_SPEED_APPS += SystemUI
 
-# Treble properties
+# Force using the following regardless of shipping API level:
+#   PRODUCT_TREBLE_LINKER_NAMESPACES
+#   PRODUCT_SEPOLICY_SPLIT
+#   PRODUCT_ENFORCE_VINTF_MANIFEST
+#   PRODUCT_NOTICE_SPLIT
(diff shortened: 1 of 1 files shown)


jerpelea:
updates?

ix5:
I've been running with locally for months now. It doesn't change any functionality, just brings all devices in line with the same VNDK packages.

jerpelea:
then why is it still under review :)

ix5:
No one bothered to test... ;)

jerpelea:
now they will :)

PR #677, opened 2020-01-24
common-prop : Add Stagefright prop

Paulbouchara:
This prop was in Pixel Experience during first bring up, add it to common as it's useful.


jerpelea:
please add a sane description :)

PR #678, opened 2020-01-27
common-prop : Add Stagefright prop

Paulbouchara:
This prop allow to prefer HW codecs on top of SW for thumbnails, saving battery, getting faster decoding and still having a fallback in case HW codecs are broken.


jerpelea:
for now it will stay open until the hw decoders are working as they should

jerpelea:
the codecs seem to behave as they should

PR #680, opened 2020-02-03
treble/vintf: Bump graphics.composer to version 2.3.

MarijnS95:
DONOTMERGE: Composer 2.3 _depends on_ 8.1 display. It'll _crash_ with 7.1, likely due to missing/unimplemented/broken functionality.

After moving to CAF 8.1 display branches the new composer can and should
be used.


jerpelea:
+1 
tested on nile, ganges, tama, kumano

kholk:
I think that you definitely want to click on the merge button right now :)))

PR #681, opened 2020-02-05
[WIP] Readd gps conf

ix5:
Might wanna overhaul it while we're at it...

Needs https://github.com/sonyxperiadev/vendor-qcom-opensource-location/pull/22 and https://github.com/sonyxperiadev/repo_update/pull/119

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -61,6 +61,10 @@ PRODUCT_COPY_FILES += \
 PRODUCT_COPY_FILES += \
     $(COMMON_PATH)/rootdir/system/etc/nfcee_access.xml:system/etc/nfcee_access.xml
 
+# GPS Configuration
+PRODUCT_COPY_FILES += \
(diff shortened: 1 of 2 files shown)


jerpelea:
why?

ix5:
I'll add the needed patches for qcom/gps

jerpelea:
link?

ix5:
https://github.com/sonyxperiadev/repo_update/blob/master/repo_update.sh#L63-L66
These need to go, and a patch to opensource location.

jerpelea:
do them and we can see

jerpelea:
but instead of adding the conf back here you should update the one from the gps repo otherwise it will create conflicts to the downstream projects

ix5:
https://github.com/omnirom/android/blob/1bcf6a6f5c082107c29977555bbd123fbcf7b47c/remove.xml#L111
https://github.com/omnirom/android_hardware_qcom_gps/tree/android-10/
SonyAOSP isn't on 10 yet afaics https://github.com/SonyAosp/platform_hardware_qcom_sdm845_gps

We're the only ones still applying those reverts on Q.

PR #682, opened 2020-02-07
[master] Audio bump to 5.0

ix5:
And also add some bt packages

Audio HAL also might need updates to `mixer_paths` https://github.com/ExplosiveLobster/device_sony_suzu/commit/535bb5dc7df965b69a823dd3b8ba292ec7224813

### DIFF
--- a/common-packages.mk
+++ b/common-packages.mk
@@ -15,6 +15,8 @@
 # Audio
 PRODUCT_PACKAGES += \
     audio.a2dp.default \
+    audio.bluetooth.default \
+    audio.hearing_aid.default \
     audio.r_submix.default \
     audio.usb.default \
     libaudio-resampler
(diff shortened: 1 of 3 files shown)


kholk:
Felix, please, remove the draft status from this PR.
We need audio-5.0 on SM8150!

MarijnS95:
@kholk I'm sorry to disappoint you, but I run 5.0 and as you know we both have the exact same audio issues on SM8150 :wink: 

Unfortunately, back to the drawing board it is :disappointed:

kholk:
We still need 5.0 on SM8150: all devices having this chip (and SM6150, and the new ones) do
come with this: switching won't hurt anyone.

Unless you tell me that there are regressions.

MarijnS95:
No regressions, I run this for months concurrently with the new A2DP offloading but it doesn't change anything for the issues we are facing now.

So by all means merge this (unless @ix5 wants to alter the commits), but don't expect magic to happen.

ix5:
Note to self: Test `IBluetoothAudioProvidersFactory` especially on legacy devices

PR #684, opened 2020-02-12
[COMMON] init.recovery: Refresh recovery USB configuration for k4.14 Android 10

kholk:
Refresh the USB configuration in order to get ADB sideloading
back up.

TEST:
SoMC Tama Akatsuki: ADB sideload working, USB composition switch OK (Recovery)

### DIFF
--- a/rootdir/init.recovery.common.rc
+++ b/rootdir/init.recovery.common.rc
@@ -1,7 +1,4 @@
 on init
-    wait /dev/block/platform/soc/${ro.boot.bootdevice}
-    symlink /dev/block/platform/soc/${ro.boot.bootdevice} /dev/block/bootdevice
-
     # USB setup
     mkdir /config 0770 shell shell
     mount configfs none /config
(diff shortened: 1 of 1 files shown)


Paulbouchara:
Tested working on SoMC Pioneer and SoMC Akatsuki and it's working fine too

MERGE MERGE

ix5:
Remove this line then https://github.com/sonyxperiadev/device-sony-common/blob/d1db224f8e6819028821bf5a94fe3354aa0b5a03/rootdir/init.recovery.common.rc#L32

kholk:
@ix5 Noupe! That line is required for a paranoid fallback to FunctionFS and in that case we still want to advertise Sony as vendor.....

ix5:
Ehm... the "paranoid" config triggers:
`property:sys.usb.config=adb && [...]`,
whereas your separate `idVendor` write triggers:
`on property:sys.usb.config=*` == "on any `sys.usb.config`.

So you're already always writing the `idVendor`.

But I guess in a race, it could happen that the `idVendor` write from the wildcard trigger is written *before* the functionfs is switched to adb, requiring another `idvVendor` write. So your paranoid section should stay :)

PR #685, opened 2020-02-19
common-init: add dpmQmiMgr service file.

MarijnS95:
This service is required on Kumano.

### DIFF
--- a/common-init.mk
+++ b/common-init.mk
@@ -21,6 +21,7 @@ PRODUCT_PACKAGES += \
     cdsprpcd.rc \
     cnss-daemon.rc \
     ipacm.rc \
+    dpmQmiMgr.rc \
     imsdatadaemon.rc \
     imsqmidaemon.rc \
     imsrcsd.rc \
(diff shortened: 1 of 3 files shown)


rootdir/vendor/etc/init/dpmQmiMgr.rc
```diff
@@ -0,0 +1,5 @@
+#start dpmQmiMgr service
+service dpmQmiMgr /odm/bin/dpmQmiMgr
```
ix5: `service vendor.dpmQmiMgr`

common-init.mk
```diff
@@ -21,6 +21,7 @@ PRODUCT_PACKAGES += \
     cdsprpcd.rc \
     cnss-daemon.rc \
     ipacm.rc \
+    dpmQmiMgr.rc \
```
ix5: Can the sm8150-specific stuff be guarded by TARGET_PRODUCT?

Also, for later, maybe guard the IMS stuff by `TARGET_HAS_IMS`.

kholk: Not a problem to have it on legacy targets, AFAIK.

PR #686, opened 2020-02-19
[WIP] Guard IMS and kumano init files

ix5:
Guard IMS init pkgs by `TARGET_USES_IMS`. Prevents inclusion for legacy devices.

Only copy `qti.rc` and `adpl.rc` for sm8150. These services would never start on other platforms.


jerpelea:
there should not be any issue to have them

PR #689, opened 2020-02-21
[COMMON] CommonConfig: Raise maximum possible loopback partitions

kholk:
On Android 10 we need more loopback partitions in order
to get APEX working as per design.


kholk:
@MartinX3 I'm mostly sure. Basically everyone supporting this (including Sony themselves on stock) are setting the max_part to 16 and I can see why...

Paulbouchara:
@martinx3 8 is fine for Android Q, but Android R have 12 more apex modules, so we'll have to expand it some day again, it's just to gain some time

MartinX3:
Aye, nice to know that it is fine.

PR #690, opened 2020-02-23
odm: Prevent including oss precompiled sepolicy on odm image.

MarijnS95:
The precompiled sepolicy rule file on /odm has precedence over the one
in /vendor. This is unfortunate because end-users will have to wait for
the next odm image to receive policy updates (mandatory when running
enforcing) or delete it manually, while they should instead receive
updates as soon as a vendor image is built after a fresh pull from our
sepolicy repository.

While the AOSP build system supports splitting policy _contexts_ [1]
which we take advantage of by setting BOARD_VENDOR_SEPOLICY_DIRS
exclusively, the precompiled policy (the actual rules) end up on odm [2]
when BOARD_USES_ODMIMAGE is true [3].

Set PRODUCT_PRECOMPILED_SEPOLICY to false to prevent this policy from
being included at all when an odm image is built.

[1]: https://android.googlesource.com/platform/system/sepolicy/+/android-10.0.0_r27/Android.mk#103
[2]: https://android.googlesource.com/platform/system/sepolicy/+/android-10.0.0_r27/Android.mk#837
[3]: https://android.googlesource.com/platform/system/sepolicy/+/android-10.0.0_r27/Android.mk#264

Test: m SONY_BUILD_ODM=1 odmimage, sepolicy folder is gone after this
      change. Sepolicy remains on vendor when SONY_BUILD_ODM is false.

---

@jerpelea This may or may not work out for you, depending on how the odm image is built. If you have set `SONY_BUILD_ODM` in customization and build vendor images with that this will break the policy entirely as it'll not end up on vendor either. But if you temporarily set it (eg. `m SONY_BUILD_ODM=1 odmimage`) this will work just fine.


jerpelea:
not needed I fixed it on v4 and reuploaded

PR #691, opened 2020-02-23
Remove HWUI properties

aviraxp:
* Useless after Oreo.


common-prop.mk
```diff
@@ -70,20 +70,6 @@ PRODUCT_PROPERTY_OVERRIDES += \
 PRODUCT_PROPERTY_OVERRIDES += \
     debug.hwui.use_buffer_age=false
```
ix5: Why not remove this as well?

aviraxp: This is still used.

PR #699, opened 2020-03-02
rootdir: vendor: move audiopd to own file

oshmoun:
Guard the separated service with a flag instead of enabling at boot time via prop.


oshmoun:
Needs to have the flag defined for platforms that need it.
Also it's untested but should hint at the general idea.

PR #701, opened 2020-03-06
General cleanup

ix5:
Does this spark joy? I hope this sparks you for you!

---

### hardware: ssc: Convert .mk to .bp
It's cleaner and faster to parse.

### rootdir: Correct km4 init module name

The module is only shipping an init file, not an executable. Clear up this confusion my appending .rc to its module name.

### rootdir: rdclean: Remove duplicate guard

The module will only be added for userdebug/eng builds in common.mk anyway.

### rootdir: Make PLATFORM guards more logical

Having this makes little sense:
```
if (A && B && C)
if (B && C)
<add-stuff>
endif
<add-more-stuff>
endif
```

Rather, change it to:
```
if (B && C)
<add-stuff>
endif

if (A && B && C)
<add-more-stuff>
endif
```

### rootdir: Drop msm8916 from guard for qmuxd.rc

This platform is not included in SODP Q.

### rootdir: Convert .mk to .bp

Since soong doesn't know about templated module names, `init.$(TARGET_DEVICE).rc` and `init.recovery.$(TARGET_DEVICE)` stay as .mk.

Also, any module that is conditionally included on `TARGET_*` vars is left as-is.

### rootdir: sscrdpcd.rc: Only for sdm845,sm8150

The service is not used by sdm660 and msm8998.

---

## TODO

- `health: vintf_fragments for manifest`
- `rootdir: Move guards to common-init`


common-init.mk
```diff
@@ -55,10 +91,13 @@ PRODUCT_PACKAGES += \
...
-# Common init scripts
+ifeq ($(WIFI_DRIVER_BUILT),qca_cld3)
 PRODUCT_PACKAGES += \
-    init.qcom.modem.sh \
-    init.qcom.adspstart.sh \
-    init.qcom.cdspstart.sh \
-    init.qcom.ipastart.sh \
-    init.qcom.slpistart.sh
+    cnss-daemon.rc \
+endif
```
MarijnS95: You might want to take a look at these two lines :stuck_out_tongue_winking_eye:

hardware/time-services/time_genoff.h
```diff
@@ -1,104 +0,0 @@
-/*
```
ix5: @jerpelea do you use this for your proprietary builds?

common-init.mk
ix5: Nothing to see here... ;)

hardware/firmware/Android.mk
```diff
@@ -1,14 +0,0 @@
...
-LOCAL_PATH := $(call my-dir)
-
-include $(CLEAR_VARS)
-LOCAL_MODULE := firmware_folders
-LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)
-
-LOCAL_POST_INSTALL_CMD := mkdir -p \
-    $(TARGET_OUT_VENDOR)/dsp \
-    $(TARGET_OUT_VENDOR)/firmware_mnt \
-    $(TARGET_OUT_VENDOR)/bt_firmware
```
MarijnS95: This commit makes no sense to me. These are the _folders in `/vendor`_ where the partitions will be mounted. The `BOARD_ROOT_EXTRA_SYMLINKS` in CommonConfig.mk _link to these folders_, not the other way around.

ix5: Dropped

ix5:
This PR is dependent on a reorganization of the device makefiles, defining variables earlier in lowercase `device.mk`-inherited makfiles.

common-init.mk
```diff
@@ -12,39 +12,75 @@
...
     per-service.rc \
-    qmuxd.rc \
     qrtr.rc \
     qseecom.rc \
     rmt_storage.rc \
     sct_service.rc \
     sensors.rc \
-    sscrpcd.rc \
-    tad.rc \
-    ta_qmi.rc \
```
MarijnS95: Notice that `ta_qmi.rc` has disappeared completely.

rootdir/Android.bp
```diff
@@ -0,0 +1,286 @@
+cc_prebuilt_binary {
+    name: "rdclean.sh",
+    srcs: ["vendor/bin/rdclean.sh"],
+    init_rc: ["vendor/etc/init/rdclean.rc"],
+    vendor: true,
+}
+
+prebuilt_etc {
+    name: "android.hardware.keymaster@4.0-service-qti",
```
MarijnS95: Note that the `.rc` postfix disappeared here. This was previously controlled by the `_STEM` in the Makefile.

ix5:
> sscrpcd should not be added to sdm660 adn msm8998
> this service is used only by the sdm845 and sm8150

As it currently stands, sscrpd is included for `msm8998` and later
`ifneq ($(filter sdm660 msm8998 sdm845 sm8150,$(TARGET_BOARD_PLATFORM)),)`
https://github.com/sonyxperiadev/device-sony-common/blob/master/rootdir/Android.mk#L320-L363

I can remove the init though.

> do not remove the time services

Ok

jerpelea:
> 
> 
> > sscrpcd should not be added to sdm660 adn msm8998
> > this service is used only by the sdm845 and sm8150
> 
> As it currently stands, sscrpd is included for `msm8998` and later
> `ifneq ($(filter sdm660 msm8998 sdm845 sm8150,$(TARGET_BOARD_PLATFORM)),)`
> https://github.com/sonyxperiadev/device-sony-common/blob/master/rootdir/Android.mk#L320-L363
> 
> I can remove the init for 660 and 8998 though.
> 
> > do not remove the time services
> 
> Ok

please do the blobs are not prezent on those devices

ix5:
Pushed with less breaking changes

ix5:
This pulls in https://github.com/sonyxperiadev/device-sony-common/pull/734 as well

MarijnS95:
@ix5 Can you pull in https://github.com/sonyxperiadev/device-sony-common/pull/735 too?

ix5:
Rebased on top of latest changes

jerpelea:
cherry-picked

PR #704, opened 2020-03-11
[q-mr1-legacy] Move kernel path to common config

chris42:
Following changes of @ix5 to fix build in legacy devices.

Platform will need change in platform.mk. See [Loire for example](https://github.com/sonyxperiadev/device-sony-loire/commit/678af16ec3d2283e0a88da9130d623d60cffa6b2)

Should be merged together with [kernel/PR2216](https://github.com/sonyxperiadev/kernel/pull/2216)

I am currently testing this for Kugo/Loire. Other device/platform tests would be nice.


ix5:
Thanks for PR-ing this!

PR #707, opened 2020-03-27
treewide: Replace COPY_HEADERS

ix5:
LOCAL_COPY_HEADERS and BUILD_COPY_HEADERS headers have been deprecated in Q and will throw an error in R.

While we're at it, convert the affected modules to blueprint.

For the propietary parts, inclusion of the new `time_services_headers` header module  will be necessary.


MarijnS95:
Tested on a Q Griffin build, and is necessary for R builds. @jerpelea MERGEME!! (After you have applied the necessary changes to the odm build, of course.)

PR #712, opened 2020-04-25
[COMMON] common-prop: Temporarily disable audio@5.0 bluetooth HAL

MarijnS95:
Audio v5 includes a new HAL and way to register bluetooth endpoints
(bluetooth.audio@2.0) and their transmission, which is used by newer
bluetooth audio libs. This obviously requires different audio policy to
route BT streams to said libs, which has not been included in the V5
bump.
Instead of switching that around, allow me to finalize A2DP offloading
(finally) and PR the entire, proper switch at once. Meanwhile, disable
this new way of registering endpoints.

### DIFF
--- a/common-prop.mk
+++ b/common-prop.mk
@@ -124,6 +124,10 @@ PRODUCT_PROPERTY_OVERRIDES += \
     ro.lmk.kill_timeout_ms=100 \
     ro.lmk.use_minfree_levels=true
 
+# Bluetooth Audio
+PRODUCT_PROPERTY_OVERRIDES += \
+    persist.bluetooth.bluetooth_audio_hal.disabled=true
+
 # Property to enable user to access Google WFD settings.
 PRODUCT_PROPERTY_OVERRIDES += \
     persist.debug.wfd.enable=0


kholk:
... Or just finalize it without reverting back stuff?

MarijnS95:
@kholk There has been too much time spent arranging things across the place. That has finally been settled; all that's necessary is a focused afternoon coding session from me and the product is here.

That said there's no real "reverting" here. This is simply disabling the new bluetooth.audio@2.0 registration HAL until I PR the entire thing with appropriate audio policy/routing instead of rebasing and sorting that out to go there partly. Of course, I could have done that in the time it took me to write a rant instead :no_mouth:.

kholk:
Ranting over people trying to get stuff working (and actually getting that stuff to work) is counter productive and doesn't help anyone.

Your PR, as you know already, also, contains basically zero explanation of what is happening and 99% rants: please reformulate the description keeping a professional attitude.

Also, please organize that btaudio thingy: if you are able to do it in just some days, it makes no sense to disable stuff to re-enable after hours.

MarijnS95:
> contains basically zero explanation [...] please reformulate the description keeping a professional attitude.

1/3rd of it actually perfectly explains why this is a problem. And I had the same idea of reformulating right before your comment, so I guess we're on the same page. It was actually **me who gave the [go ahead](https://github.com/sonyxperiadev/device-sony-common/pull/682#issuecomment-619029798)** anyway without fully realizing there would be a problem, only predicting _something_ hence the followup comment.

> Also, please organize that btaudio thingy: if you are able to do it in just some days, it makes no sense to disable stuff to re-enable after hours.

IMO it is totally fine to disable it for now. I hope it takes a few days, but it might be weeks (it has already taken a year)...
Meanwhile, users building and flashing this are left without audio on their Bluetooth headset or speakers, which is a much worse thing to do. As I need to change this and other props anyway, it's totally fine to get this in right now.

MarijnS95:
This will be superseded by adjusted policies that are constructive to a2dp offloading at the same time.

PR #717, opened 2020-04-29
[R] packages: Add protobuf vendorcompat lib

ix5:
Our blobs are compiled against `libprotobuf-full-cpp.so`, but starting in R, Android is using a versioned naming approach, e.g. `libprotobuf-cpp-full-3.9.1.so`.

See https://r.android.com/1109518

The entry in `public.libraries.txt` is needed to allow the linker to find and use the lib outside the VNDK.

See https://source.android.com/devices/tech/config/namespaces_libraries

### DIFF
--- a/common-packages.mk
+++ b/common-packages.mk
@@ -121,6 +121,11 @@ PRODUCT_PACKAGES += \
     libandroid_net \
     libprotobuf-cpp-full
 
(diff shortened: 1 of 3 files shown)


ix5:
This is ready to merge into R/master.

PR #719, opened 2020-04-30
vintf: Align NFC

ix5:
### Align INfc version, fix INxpNfc version

We are shiping the 1.2 nxp nfc service now.
The vendor.nxp version was wrong, there is no @1.1 interface as of now.

### DIFF
--- a/vintf/manifest.xml
+++ b/vintf/manifest.xml
@@ -156,7 +156,7 @@
     <hal format="hidl">
         <name>android.hardware.nfc</name>
         <transport>hwbinder</transport>
-        <version>1.1</version>
+        <version>1.2</version>
         <interface>
             <name>INfc</name>
             <instance>default</instance>
--- a/vintf/vendor.nxp.nfc.interfaces.xml
+++ b/vintf/vendor.nxp.nfc.interfaces.xml
@@ -2,7 +2,7 @@
     <hal format="hidl">
         <name>vendor.nxp.nxpnfc</name>
         <transport>hwbinder</transport>
-        <version>1.1</version>
+        <version>1.0</version>
         <interface>
             <name>INxpNfc</name>
             <instance>default</instance>


kholk:
This is the kind of stuff that makes me want to break everything around me and start screaming like a crazy man.

Of course my Griffin was tested with this stuff.
Of course I'm dumb and I forgot to push it. 

Of course this shouldn't have the DONOTMERGE tag............

ix5:
Don't merge yet because somehow the secure element instance that is reported by `lshal` differs.
```
griffin:/ # lshal | grep -ie nfc -e nxp -e secure_e
DM    Y android.hardware.nfc@1.0::INfc/default                                                    0/1        741    2704 591
DM,FC Y android.hardware.nfc@1.1::INfc/default                                                    0/1        741    2704 591
FC    Y android.hardware.nfc@1.2::INfc/default                                                    0/1        741    2704 591
FC    Y android.hardware.secure_element@1.0::ISecureElement/SIM1                                  0/2        916    591
FC    Y android.hardware.secure_element@1.0::ISecureElement/SIM2                                  0/2        925    591
DM    Y vendor.nxp.nxpnfc@1.0::INxpNfc/default                                                    0/1        741    591
```

So it seems the SE service also fails to bring up the `eSE1` instance.

kholk:
Strange, it was working for me when I checked...
Anyway, it should be safe to push the nxpnfc and nfc commits, as the issue that you see is only on the secure element one, isn't it?

In that event, perhaps push the nfc-only ones separately so that we can get them up sooner than later?

ix5:
Split out the SE stuff into https://github.com/sonyxperiadev/device-sony-common/pull/720

MarijnS95:
> Strange, it was working for me when I checked...

And, of course it would :wink: 

https://source.android.com/devices/architecture/vintf/match-rules#DRM

Minor versions are allowed to be equal or larger than the manifest. What happens with `INxpNfc` I don't know, perhaps an optional and in this case unused interface, or the vintf enforcer sees it through...

PR #721, opened 2020-05-03
hw: qcom: do not explictly include hals

oshmoun:
Display, media and ipa hals are now automatically included by the build system.
Explictly including them leads to double module definitions.
Additionally, add a brief documentationfor QCOM_NEW_MEDIA_PLATFORM.

### DIFF
--- a/hardware/qcom/Android.mk
+++ b/hardware/qcom/Android.mk
@@ -5,6 +5,9 @@ else
 # TARGET_BOARD_PLATFORM specific featurization
 QCOM_BOARD_PLATFORMS += msm8952 msm8996 msm8998 sdm660 sdm845 sm8150
 
+# Some supported platforms need a different media hal
+# This list selects platforms that should use the latest media hal
+# All other platforms automatically fallback to the legacy hal
 QCOM_NEW_MEDIA_PLATFORM := sdm845 sm8150
 
 #List of targets that use video hw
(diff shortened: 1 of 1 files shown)


hardware/qcom/Android.mk
```diff
@@ -15,14 +18,14 @@ MASTER_SIDE_CP_TARGET_LIST := msm8996 msm8998 sdm660 sdm845 sm8150
 
 audio-hal := hardware/qcom/audio
 gps-hal := hardware/qcom/gps/sdm845
-ipa-hal := hardware/qcom/data/ipacfg-mgr/sdm845
+ipa-hal := vendor/qcom/opensource/data/ipacfg-mgr/sdm845
 
-display-hal := hardware/qcom/display/sde
+display-hal := vendor/qcom/opensource/display/sde
```
ix5, line 23: Why not just move it do `opensource/display`?
We only chose the `sde` subfolder back then for... I think it was to differentiate from `msmfb`?

oshmoun: good suggestion, done

hardware/qcom/Android.mk
```diff
@@ -35,9 +38,6 @@ TARGET_KERNEL_VERSION := $(SOMC_KERNEL_VERSION)
 
 include device/sony/common/hardware/qcom/utils.mk
 
-include $(display-hal)/Android.mk
 include $(call all-makefiles-under,$(audio-hal))
-include $(call first-makefiles-under,$(ipa-hal))
 include $(call all-makefiles-under,$(gps-hal))
-include $(call all-makefiles-under,$(media-hal))
```
MarijnS95, line 42: If these are removed, why not drop the now-unused variables above too?

oshmoun: because i saw there are other seemingly unused variables in there prior to this change. that made me assume these are somehow used elsewhere, and no i did not check :D

ix5: The lowercase vars aren't used anywhere else IIRC, but the uppercase vars are being referred from some HALs theselves.

oshmoun: sure can clean them up then, will do

PR #722, opened 2020-05-15
common-prop: Disable PartyCrasher - don't reboot to recovery for wipe

MarijnS95:
Seriously. A boot-to-recovery instead of an _at least partially working_
device? /shrug.

The intention is to clean the device and at last resort request the user
to wipe their phone in an attempt to recover the device.

In our case this is hardly ever a solution: rather we want to get to the
bottom of an issue/crash and resolve it, as we have the ability to do so
on an open platform.

I have personally not seen these "recovery attempts" yet, rather an
immediate reboot to recovery, hiding actual issues at hand (90% of the
time sensors missing -> transpower crashing -> taking the core phone
process with it).

See the docs for more info:
https://source.android.com/devices/tech/debug/rescue-party

---

Partycrasher is a [common](https://github.com/sonyxperiadev/bug_tracker/issues/471#issuecomment-535959689) [way](https://github.com/sonyxperiadev/bug_tracker/issues/580) [to](https://github.com/sonyxperiadev/bug_tracker/issues/351#issuecomment-531491110) [hide](https://github.com/sonyxperiadev/bug_tracker/issues/572#issuecomment-622357146) issues from us. Let's disable it and save ourselves the hassle sorting through logs and figure out the device isn't actually rebooting due to hardware errors.

One reason to not merge this: Issues might be harder to spot. While crashing `com.android.phone` _should_ show on the screen, I don't recall seeing it nor did any of the report indicate such. That means the device drains a ton of energy, and when you're finally outside and need to make an emergency call...

### DIFF
--- a/common-prop.mk
+++ b/common-prop.mk
@@ -215,3 +215,8 @@ endif
 # Reduce cost of scrypt for FBE CE decryption
 PRODUCT_PROPERTY_OVERRIDES += \
     ro.crypto.scrypt_params=15:3:1
+
+# Disable PartyCrasher: No need to reboot to recovery
+# and request a wipe when something is broken.
+PRODUCT_PROPERTY_OVERRIDES += \
+    persist.sys.disable_rescue=1


ix5:
We want to keep the party going! :tada:
No lame party crashers for us.

stefanhh0:
> While crashing com.android.phone should show on the screen, I don't recall seeing it nor did any of the report indicate such. That means the device drains a ton of energy, and when you're finally outside and need to make an emergency call...

Sometimes a box appears showing that the phone subsys crashed, sometimes not. Anyway if one can not do a call, reboot is the most common and well known workaround... Actually, it is "the" workaround since we know and use computers, right?

MarijnS95:
> Sometimes a box appears showing that the phone subsys crashed, most of the times not.

That's rather discouraging: I would hope it alerts the user every time on a boot that something _important_ is not right.

> Anyway if one can not do a call, reboot is the most common and well known workaround... Actually, it is "the" workaround since we know and use computers, right?

This would be a specific case where the user is not (temporarily) able to make a call. In this instance it should be impossible right after boot, and stop crashing+restarting after a few minutes.

Better hope users test core functionality of their device before taking it out, especially with bringups and changes going on to support new platforms all the time. Besides that we are (I am, at least :stuck_out_tongue_winking_eye:) known to break things unintentionally (how else can we, ~3-5 people working in their free time to keep ~19 devices alive? Luckily all the testers/users out there catch it so we can tackle it quickly :+1: )

> Regarding energy waste resp. drained device. Many have powerbanks and most of the time there are many people outside having a phone and you can always ask, don't be too shy :-) Well one could think of cases where it safe lives but not too likely. I don't believe that many people are going to die because party crasher has been removed from sony aosp devices?!

Oh - I meant that the only symptom is a power-hungry device... Then when you need to place a call you find out the entire telephony stack is MIA...
(Though in this specific case, iirc it "gives up" at some point and you'll still be able to open the dialer and place a call. Don't quote me on that / rely on it though :wink:)

stefanhh0:
You are all are doing a great job, given with what resources you are dealing, how big the project is, that it is mostly done on a free time basis. I still think it would be great if Sony would just offer a job opportunity for those people that are very involved here and doing such a great job, you deserve it  :-) I guess that some of the development goes back to the stock, right? Especially what you guys are developing on kernel level... so I guess Sony benefits from your work here as well.

> (Though in this specific case, iirc it "gives up" at some point and you'll still be able to open the dialer and place a call. Don't quote me on that / rely on it though :wink:) 

If the phone subsystem does not recover eventually, it crashes at a later point in time, sometimes it recovers and remains stable until the next crash... In fact when I have my workaround active not letting the phone fall asleep it stays stable for as many hours as it is plugged most of the times.

I am running my phone already a couple of days without party crasher and love it. :heart:

MarijnS95:
Hehe, thanks :grin:. This wasn't a request for "oh please motivate me we're so sad" even though it sounds like one; just an objective explanation (to other readers) why it is tough to keep devices afloat all the time. As discussed previously "simply slapping a CI on here" would cover only 0.0005% of the mistakes we make here: by far the most is only found on-device. That is incredibly difficult, costly and time-consuming to set up and keep running if it weren't for all our testers out here :smile: 

As far as I know nothing goes back into stock. They have their own teams that are much larger than us, and obviously manufacturer support for all the hardware they use. Though they do reap the benefits from fixes and improvements we (or rather, @ix5) submit to AOSP Gerrit, and perhaps kernel patches in the future (even if for legacy SoCs to begin with).

Looks like two issues are crossing over here: a modem crash (as seen in the sonyxperiadev/bug_tracker#580) and sensors not starting -> takes down transpower -> takes down AOSP phone app/stack. The latter one seems to "resolve itself" after a while as "something" stops probing sensors.

As long as we have proper means to show something is wrong (the crash shown in UI) this patch is safe to merge and saves us from going crazy over unnecessary "USB related" reboots.

PR #729, opened 2020-05-22
[VERIFYME] tftp_symlinks: Use updated persist and fw targets

ix5:
We should be trying to get rid of the rootfs `/firmware` and `/persist` compatibility symlinks.
That entails actually not using them any more...


MarijnS95:
Change looks sane to me :ok_hand: 

No issues encountered on Bahamut nor Akatsuki, all RFS stuff seems to function normally (as expected, of course).

ix5:
Thanks for testing!

PR #732, opened 2020-05-23
[DNM] location: update to gnss 2.0 and caf 8.1 gps hal

oshmoun:


hardware/qcom/Android.mk
```diff
@@ -17,7 +17,7 @@ MSM_VIDC_TARGET_LIST := msm8952 msm8996 msm8998 sdm660 sdm845 sm8150
 MASTER_SIDE_CP_TARGET_LIST := msm8996 msm8998 sdm660 sdm845 sm8150
 
 audio-hal := hardware/qcom/audio
-gps-hal := hardware/qcom/gps/sdm845
+gps-hal := vendor/qcom/opensource/gps
```
MarijnS95, line 20: The sole use of `$(gps-hal)` below is dropped. Does it still make sense to set this?

While at it, `ipa-hal`, `media-hal` and `audio-hal` were missed in the previous cleanup, too. Only `audio-hal` remains with an explicit include.

jerpelea:
the explicit declaration is used by vendor to locate the needed folders 
please update them all to the new locations

oshmoun:
removed unneeded variables

PR #733, opened 2020-05-25
[COMMON] [URGENT] common-odm: Remove OSS components

kholk:
The ims-ext-common, qti-telephony-wrapper,
qti-telephony-common are open source
components, found in vendor-qcom-opensource-telephony.

### DIFF
--- a/common-odm.mk
+++ b/common-odm.mk
@@ -173,9 +173,6 @@ PRODUCT_PACKAGES += \
     vendor.qti.latency@2.0 \
     QtiTelephonyService \
     QtiTelephonyServicelibrary \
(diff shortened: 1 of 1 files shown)


jerpelea:
please move them in the common-packages

kholk:
Alin they're already there too....
https://github.com/sonyxperiadev/device-sony-common/blob/master/common-packages.mk#L70

Paulbouchara:
https://github.com/sonyxperiadev/device-sony-common/blob/master/common-packages.mk#L70

They're already there

jerpelea:
it is my fault for the double inclusion

PR #739, opened 2020-05-27
[DNM] init: Import charger from coral

ix5:
- Move charger into its own init file
- Add charger.rc to packages
- Fix the location (the symlink from root/charger points to
  /system/bin/charger)
- Allow powersaving when in charger mode, drop root user.

### DIFF
--- a/common-init.mk
+++ b/common-init.mk
@@ -20,6 +20,7 @@ PRODUCT_PACKAGES += \
     adsprpcd.rc \
     audiopd.rc \
     cdsprpcd.rc \
+    charger.rc \
     cnss-daemon.rc \
     ipacm.rc \
     dataqti.rc \
--- a/rootdir/Android.bp
+++ b/rootdir/Android.bp
@@ -60,6 +60,13 @@ prebuilt_etc {
 }
 
 prebuilt_etc {
+    name: "charger.rc",
+    src: "vendor/etc/init/charger.rc",
+    sub_dir: "init",
+    vendor: true,
+}
+
+prebuilt_etc {
     name: "ipacm.rc",
     src: "vendor/etc/init/ipacm.rc",
     sub_dir: "init",
--- a/rootdir/vendor/etc/init/charger.rc
+++ b/rootdir/vendor/etc/init/charger.rc
@@ -1,0 +1,37 @@
+on charger
+    # To mount /persist for health HAL inside charger
+    trigger fs
+
+    start system_suspend
+    start vendor.rmt_storage
+    start vendor.mlog_qmi_service
+    start vendor.irsc_util
+    # Should not be needed
+    #start vendor.qrtr-ns
+    #start vendor.tftp_server
+    start vendor.per_mgr
+
+on property:init.svc.vendor.charger=running
+    stop vendor.qseecomd
+    stop vendor.gatekeeper-1-0
+    # Legacy devices
+    stop vendor.keymaster-3-0
+    # KM4+ devices
+    stop vendor.keymaster-4-0
+    write /sys/class/typec/port0/port_type sink
+    write /sys/module/lpm_levels/parameters/sleep_disabled 0
+    setprop sys.usb.configfs 1
+
+# Offline charger
+service vendor.charger /system/bin/charger
+    class charger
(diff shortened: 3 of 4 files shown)


MarijnS95:
This conflicts with your `.bp` conversion, would you mind to rebase it? In any case it seems to work fine on Discovery RoW and Griffin DSDS, killing unnecessary services seems like a good saver even if the gain will be negligible at best.
I have yet to enable adb in charger mode and see if it actually behaves properly, though.

ix5:
> This conflicts with your `.bp` conversion, would you mind to rebase it?

Anything for my favourite Flying Dutchman. Done.

> In any case it seems to work fine on Discovery RoW and Griffin DSDS, killing unnecessary services seems like a good saver even if the gain will be negligible at best.

I'm more worried about heat buildup preventing faster charging, and if we can get the system to basically go into winter hibernation mode, all the better.

> I have yet to enable adb in charger mode and see if it actually behaves properly, though.

Yeah there might also be some weird stuff going on depending on the bootloader.
Also, I'd ideally like to keep the charger binary inside the ramdisk, even if elgoog dissuades you from doing so.

MarijnS95:
> Anything for my favourite Flying Dutchman. Done.

Dank!
 
> I'm more worried about heat buildup preventing faster charging, and if we can get the system to basically go into winter hibernation mode, all the better.

Totally, heat buildup got lost in my comment while refactoring. There's definitely not going to be much if any gain at all in charging speed, but the tiny bit of lost heat (from potentially looping services) might be worth it. That said they are all binder and block indefinitely on a poll or ioctl, meaning no cycles wasted in userspace nor kernel.
 
> Yeah there might also be some weird stuff going on depending on the bootloader.

Which basically means we have to check all devices carefully?

> Also, I'd ideally like to keep the charger binary inside the ramdisk, even if elgoog dissuades you from doing so.

Being able to charge without working /system or /vendor seems like a solid thing to do, as long as we have space.

ix5:
> Which basically means we have to check all devices carefully?

Sadly yes. I don't want to break this critical functionality for anyone.

Also, not every device supports e.g. booting directly from off-mode charging via `ro.enable_boot_charger_mode`, could also be interesting.
 
> Being able to charge without working /system or /vendor seems like a solid thing to do, as long as we have space.

We should, since so much stuff got moved out of the initramdisk with SAR (system-as-root) and 2SI (two-stage-init).

Buuuut since _everything_ in Android world is now binderized, the `charger` binary - which is actually our old "friend" `healthd` in disguise, relies on [this collection of modules](https://android.googlesource.com/platform/system/core/+/refs/tags/android-10.0.0_r35/healthd/Android.mk#161), linked.

Dream -> poof.

MarijnS95:
> Sadly yes. I don't want to break this critical functionality for anyone.

Totally. And we should regardless: I tried Mermaid and the panel turns on without showing anything (as expected) but gets hot near the CPU and doesn't turn off after being unplugged. Discovery is fine however. (with and without this PR - something different is going on)

> Also, not every device supports e.g. booting directly from off-mode charging via `ro.enable_boot_charger_mode`, could also be interesting.

off-mode charging? Is this holding the power button while it is in charger mode and have it exit gracefully, booting into Android without reboot?
I just held the button on Griffin and it froze up a while before panicking and rebooting (into charger, no way to recover pstore unless we kick off adb). Again: we have to check all devices :joy: 

> We should, since so much stuff got moved out of the initramdisk with SAR (system-as-root) and 2SI (two-stage-init).

Yup, should be fine.

> Buuuut since _everything_ in Android world is now binderized, the `charger` binary - which is actually our old "friend" `healthd` in disguise, relies on [this collection of modules](https://android.googlesource.com/platform/system/core/+/refs/tags/android-10.0.0_r35/healthd/Android.mk#161), linked.
> 
> Dream -> poof.

Hmm, those are anyway core libraries. And it likely only needs the `health@2.0` interface lib to satisfy the linker, it won't actually be called by system :neutral_face:. Dreams destroyed.

ix5:
Splitting charger into its own rc is kinda stupid since in charger mode, charger.rc might not yet be available. Meh

PR #740, opened 2020-05-27
[DNM] overlay: Align PinnerService files with coral

ix5:


ix5:
Ok, half of the pinned files don't even exist because we only pre-opt System Server and boot image stuff.

See `WITH_DEXPREOPT_BOOT_IMG_AND_SYSTEM_SERVER_ONLY` in [CommonConfig.mk](https://github.com/sonyxperiadev/device-sony-common/blob/1874b3249e8feb659799130324151f55868af30d/CommonConfig.mk#L121).

MarijnS95:
@ix5 This is superseded by #784 with a fresh configuration, all files exist on Griffin and should likely exist on other platforms too (at least `/apex`).

I don't think we need `config_apexBootImagePinnerServiceFiles` (or `config_jitzygoteBootImagePinnerServiceFiles` as it were on R) as we're not running [a special boot image](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/core/java/com/android/server/PinnerService.java;l=272-284?q=config_jitzygoteBootImagePinnerServiceFiles).

ix5:
Closing in favour of https://github.com/sonyxperiadev/device-sony-common/pull/784

PR #741, opened 2020-05-29
rootdir: Restore LOCAL_PATH call

ix5:
Got lost in 35c3a0fea71f6c61ec81ea7989562ae8c1d8c954
"rootdir: Convert .mk to .bp"

Still needed because else Android's build system does not
automatically make paths relative to the file unless
LOCAL_PATH is called.


MarijnS95:
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/588

This line got lost _twice_, goes to show that keeping paths always relative to the current folder is a good idea, like they _finally_ do in `.bp`.

Merge!

PR #742, opened 2020-05-29
Partially revert "CommonConfig: Enable OSS sound trigger support"

ix5:
This reverts commit eaeefad2706349531097346ed78b17cae731185f.

Reason for revert: Not all platforms *really* support sound triggers yet. Leave the `BOARD_SUPPORTS_SOUND_TRIGGER` per-platform, as it is already done.

(Real reason: 8996 does not support ST functionality and there are no blobs)

**NOTE:** This does not change any functionality, it only fixes builds for my local legacy platforms.
```
$ rg BOARD_SUPPORTS_SOUND_TRIGGER
kumano/PlatformConfig.mk
91:BOARD_SUPPORTS_SOUND_TRIGGER := true

tama/PlatformConfig.mk
91:BOARD_SUPPORTS_SOUND_TRIGGER := true

nile/PlatformConfig.mk
91:BOARD_SUPPORTS_SOUND_TRIGGER := true

ganges/PlatformConfig.mk
88:BOARD_SUPPORTS_SOUND_TRIGGER := true
```


MarijnS95:
@jerpelea Looks like `BOARD_SUPPORTS_SOUND_TRIGGER` is missing from Seine, is that intentional?

ix5:
> @jerpelea Looks like `BOARD_SUPPORTS_SOUND_TRIGGER` is missing from Seine, is that intentional?

While we're at it, might check whether yoshino can also support ST.
Afaics all platforms except tama and kumano lack the proper audio policy at the moment, though.

jerpelea:
@ix5 please add sound trigger and the appropriate policy on all

PR #747, opened 2020-06-03
TEMPORARY: liblights: Disable broken "breath" feature

MarijnS95:
Writing 1 to the breath file makes the TRILED not work at all. Disable
it until we find time to check the kernel PWM pieces.

---

@Haxk20 As requested: please test it!

As for the desync issue, please seee if you can reproduce it, and if so apply the following patch on top and let me know whether that fixes it:
https://github.com/MarijnS95/device-sony-common/commit/might-address-desync

### DIFF
--- a/hardware/liblights/Light.cpp
+++ b/hardware/liblights/Light.cpp
@@ -215,6 +215,19 @@ int Light::setLightBacklight(const LightState &state)
     return err;
 }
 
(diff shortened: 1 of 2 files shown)


MarijnS95:
After testing and reporting many configurations (https://github.com/sonyxperiadev/bug_tracker/issues/577) this is just too unreliable (across platforms) to be merged. We should scrutinize and fix the kernel driver + DT instead.

PR #749, opened 2020-06-05
[COMMON] packages: Add ModemConfig

MarijnS95:
Our new modem configuration utility, that provokes sony-modem-switcher
to send carrier-specific firmware to the modem, is finally done :)


kholk:
gogogogogogogo :))))

PR #751, opened 2020-06-07
[WIP] Init optimizations

ix5:
I mainly made this PR to spite Angelo.

### DIFF
--- a/common-prop.mk
+++ b/common-prop.mk
@@ -224,13 +224,28 @@ PRODUCT_PROPERTY_OVERRIDES += \
 # Keymaster version to differentiate between legacy, v3 and v4
 ifeq ($(TARGET_LEGACY_KEYMASTER),true)
 PRODUCT_PROPERTY_OVERRIDES += \
-    ro.vendor.keymaster.version=legacy
+    ro.vendor.keymaster.version=legacy \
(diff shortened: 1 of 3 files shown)


ix5:
Requires changes to fstab in platform gits.

oshmoun:
instead of hardcoding hal versions, maybe consider adding and using props or are props unavailable at this init stage?

ix5:
> instead of hardcoding hal versions, maybe consider adding and using props or are props unavailable at this init stage?

I've added what you've asked for to evaluate for others, but please read my opinion on that in https://github.com/sonyxperiadev/device-sony-common/pull/751/commits/1a796b881adc5583d30dcbf2d7cd58966ad79f49

MarijnS95:
And as we discussed this property doesn't nearly cover all cases:
- Pacakges in common-treble.mk (only linked here by means of a comment)
- Vintf
- AOSP initrc service naming that may change or be updated

Trying to make this generic and "future proof" (which I was initially all for) is merely shooting ourselves in the foot.

chris42:
Regarding Loire/Kugo this is a complicated PR. The regular legacy build for Loire runs on the q-mr1-legacy branch, which hasn't seen the changes over the last time. 
Hence this PR and init optimizations are on a different path which will not affect Loire.

PR #753, opened 2020-06-08
[COMMON] Additional configuration for new devices and new features

kholk:
The ADSP on some targets have a Sensors PDR Service: add the
init script in order to enable the transport endpoint to the
tms/servreg service, located on sensors_pdr_adsprpc.

This init script gets copied when the target device declares
TARGET_NEEDS_SENSORS_PDR := true

EDIT:
Also, push netmgr/dsi config for SM6125 and add the QSEE
TZ Storage configuration and paths, required by new devices
and new BSPs. 

Required by new SoCs.

### DIFF
--- a/common-init.mk
+++ b/common-init.mk
@@ -41,6 +41,7 @@ PRODUCT_PACKAGES += \
     qseecom.rc \
     rmt_storage.rc \
     sct_service.rc \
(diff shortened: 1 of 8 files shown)


kholk:
Pushed new commits for further configuration of new services.

PR #754, opened 2020-06-09
rootdir: init.common: Bindmount /oem earlier

ix5:
No need to wait until late-fs if `/vendor` is already available after `mount_all --early` has run.


stefanhh0:
Tested on Yoshino/Lilac, see attached logcat in:
https://github.com/sonyxperiadev/bug_tracker/issues/597#issuecomment-641696819

PR #750, opened 2020-06-07
[COMMON] [DO-NOT-MERGE-BREAKS-ALL] vendor: init: Early mounts on fs, late mounts on post-fs

kholk:
For Dynamic Partitions devices, we need to do early mounts on fs
and late mounts on post-fs, or havoc happens.

This also means that we have to wait_for_keymaster before late
mounting, as the late ones include the userdata partition, which
is encrypted.

Before merging this, we should update all the fstab of all platforms
to declare early mounts, or everything breaks.

PR #757, opened 2020-06-11
init.common: Fix service start order for FDE

ix5:
FDE devices experience race conditions because `/data` might not be available for the services started `on late-fs`.

Rather, delay until `/data` is mounted.

Tested on Kagura with `forceencrypt=footer` in fstab, and "Secure Startup"
requiring PIN entry enabled.

### DIFF
--- a/rootdir/vendor/etc/init/hw/init.common.rc
+++ b/rootdir/vendor/etc/init/hw/init.common.rc
@@ -112,11 +112,12 @@ on fs
 on late-fs
     mount none /vendor/oem /oem bind rec
 
-on post-fs
-    # Wait for qseecomd to be started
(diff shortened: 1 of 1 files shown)


MarijnS95:
Nice one, didn't understand why the previous pr/commit had to split things in two `on late-fs` blocks anyway.

ix5:
Tested with both FDE and verity, seeing no issues on kagura. But loire seems to suffer a single reboot before getting into UI.

stefanhh0:
Running this one since many days on yoshino/lilac seems to be fine. Can this one be only merged back together #750,  #754 and the platform specific changes (e.g. for yoshino https://github.com/sonyxperiadev/device-sony-yoshino/pull/157) for early mount? Or can it be merged back independently?

PR #758, opened 2020-06-13
vendor: GP FS: Drop /data/system/users/ from whitelist

MarijnS95:
This whitelist (though it does not exclude but rather INCLUDE) moves all
paths starting with one of the entries into gp_data_path to solve system
boundary issues such as files in /data/{system,misc}, see our BUGS in
sepolicy for tee [1].

Unfortunately this breaks FPC (which is one such violation) in the
process:

    [pid  8105] openat(AT_FDCWD, 0x77e677a004, O_RDONLY|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
    [pid  8105] openat(AT_FDCWD, "/data/vendor/tzstorage//data/system/users/0/fpdata/user.db.tmp", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
    [pid  8105] openat(AT_FDCWD, "/data/vendor/tzstorage//data/system/users/0/fpdata/", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
    [pid  8105] openat(AT_FDCWD, "/data/vendor/tzstorage", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = 36
    [pid  8105] close(36)                   = 0
    [pid  8105] openat(AT_FDCWD, "/data/vendor/tzstorage//data/system/users/0/fpdata/user.db.tmp", O_RDWR|O_CREAT|O_SYNC, 010700) = -1 ENOENT (No such file or directory)

No subdirectories are created beyond /data/vendor/tzstorage making the
feature "useless", at least on Tama. Solve that and we likely run into
different issues because the FPC HAL still uses the normal path.
Besides, AOSP deliberately provides this path [2] for the HAL to store
such files in. Unfortunately the bug referenced in our sepolicy is
inaccessible to the public [3].

Final note: Whoever invented gp_whitelist_count, I have no words. Either
parse all the nodes or ask the XML parser to count it for you, this is
far from the idiomatic...

[1]: https://github.com/sonyxperiadev/device-sony-sepolicy/blob/af8240afe01e4e57aab2d30d02d6c43303b8121a/vendor/tee.te#L22-L30
[2]: https://android.googlesource.com/platform/hardware/interfaces/+/android-10.0.0_r37/biometrics/fingerprint/2.1/IBiometricsFingerprint.hal#145
[3]: https://issuetracker.google.com/issues/36644492

### DIFF
--- a/rootdir/vendor/etc/gpfspath_oem_config.xml
+++ b/rootdir/vendor/etc/gpfspath_oem_config.xml
@@ -52,8 +52,7 @@ number of path entries.
 <sfs_path>
     <gp_data_path> /data/vendor/tzstorage/ </gp_data_path>
     <gp_persist_path> /mnt/vendor/persist/data/ </gp_persist_path>
-    <gp_whitelist_count> 4 </gp_whitelist_count>
-    <gp_whitelist_path> /data/system/users/ </gp_whitelist_path>
+    <gp_whitelist_count> 3 </gp_whitelist_count>
     <gp_whitelist_path> /data/misc/qsee/ </gp_whitelist_path>
     <gp_whitelist_path> /qwes </gp_whitelist_path>
     <gp_whitelist_path> /qwes/licenses </gp_whitelist_path>


kholk:
Remove the DNM, merge this.

MarijnS95:
@kholk Two questions remain:

- Will this break anything on the new platform?
- Will the remaining whitelist paths break other features (on other devices)?

The second one is for us all to answer by just using the device. Egistec should not be affected as that uses `/data/misc/egistec` while the whitelist only includes `/data/misc/egistec`.

kholk:
I'm not sure, but without this whitelist I see stuff complaining continuously on seine... I have no idea why, but qseecom tends to refuse to start without this: especially libdrmfs loading in the qsee... Without this file it fails the init and qsee goes nuts.... 

I think that it's probably even fine if we keep this file "empty", as it seems to just look for it to be present to load the whitelist.

In any case, let's just go for this PR as it is, urgently merge it, as otherwise I have broken devices on the other side.

MarijnS95:
> I'm not sure, but without this whitelist I see stuff complaining continuously on seine... I have no idea why, but qseecom tends to refuse to start without this: especially libdrmfs loading in the qsee... Without this file it fails the init and qsee goes nuts....

As I told you (and write here for posterity) I'll re-check this and see what exactly breaks without this file to see if we can minimize the contents.
 
> I think that it's probably even fine if we keep this file "empty", as it seems to just look for it to be present to load the whitelist.

Could be one of the other config options, qseecomd on current platforms look for it too but don't crap out at least.

> P.S.: why have I pushed it to common and not seine? That's because it's software related, we run the same sw on all and if it acts like this on seine, I expect the same breakage on all, of course.

That's totally understandable and expected, too. I now caught this error early, instead of going insane during the Seine FPC bringup where this path would suddenly be prepended for no good reason, requiring the usual "deep dive". We'd just shoot ourselves in the foot if the file can be used on all platforms and is generic enough, yet is only applied to a single platform "just because". We can't handle too much divergence.

PR #760, opened 2020-06-22
[DNM] hardware: qcom: do not explicitly call audio hal

oshmoun:
The aosp audio hal was removed, and replaced with a new hal under vendor/qcom/opensource
Remove the explicit call and rely on the build system to include the hal


pablomh:
Hi @oshmoun, can you please rebase?

MarijnS95:
@pablomh it looks like this PR is obsolete since @kholk inserted a guard to only include this when `TARGET_USES_AOSP_AUDIO_HAL` is true.

pablomh:
Thanks for the info, @MarijnS95. I'll drop it from my testing PR list :)

oshmoun:
oh good to know, closing this then to avoid confusion

PR #764, opened 2020-06-24
[COMMON] init: Introduce support for Sensors PDR on SDSP

kholk:
The higher tier SoCs have got a dedicated Sensors Low Power Island
(SLPI) with its own RTOS: in this case the Sensors PDR is not
located on the ADSP, but on the SDSP (SSC).

For this reason, introduce a new init definition for the latter,
change it for the former, and activate it on these conditions:
1. TARGET_NEEDS_ADSP_SENSORS_PDR is *false*
2. TARGET_NEEDS_SDSP_SENSORS_PDR is *true*.

For obvious reasons, *no* target can have a Sensors PDR on both
A- and S- DSP... That's why one force-excludes the other.
Also, to avoid issues with having two of them together, trigger
a build error in case both env vars are true, so that the
developer is warned about the impossible *stupid*  choice.

--------

The PDR init is verified working on Sony SM8150 Kumano Griffin DSDS,
but I want some feedback from @MarijnS95 @ix5 about the way I have
done the SDSP-ADSP exclusion and the enable on the sensors.rc before
merging it.

Specifically, I want to remind you that not all devices have got either one
or the other: some legacy ones have got no Sensors PDR at all.

### DIFF
--- a/rootdir/Android.mk
+++ b/rootdir/Android.mk
@@ -40,11 +40,26 @@ LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/etc/init
 include $(BUILD_PREBUILT)
 endif
 
-ifeq ($(TARGET_NEEDS_SENSORS_PDR), true)
+# Note: Sensors PDR can be on ADSP **or** on SDSP
+ifeq ($(TARGET_NEEDS_ADSP_SENSORS_PDR), true)
+ # Let's error out if the developer makes an impossible choice
+ ifeq ($(TARGET_NEEDS_SDSP_SENSORS_PDR), true)
+  $(error Activating both ADSP and SDSP Sensors PDR is wrong!!!)
+ endif
+
 include $(CLEAR_VARS)
-LOCAL_MODULE := sensorspd.rc
+LOCAL_MODULE := adsp-sensorspdr.rc
 LOCAL_MODULE_CLASS := ETC
-LOCAL_SRC_FILES := vendor/etc/init/sensorspd.rc
+LOCAL_SRC_FILES := vendor/etc/init/adsp-sensorspdr.rc
+LOCAL_MODULE_TAGS := optional
+LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/etc/init
+include $(BUILD_PREBUILT)
+
+else ifeq ($(TARGET_NEEDS_SDSP_SENSORS_PDR), true)
+include $(CLEAR_VARS)
+LOCAL_MODULE := sdsp-sensorspdr.rc
+LOCAL_MODULE_CLASS := ETC
+LOCAL_SRC_FILES := vendor/etc/init/sdsp-sensorspdr.rc
 LOCAL_MODULE_TAGS := optional
 LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/etc/init
 include $(BUILD_PREBUILT)
--- a/common-init.mk
+++ b/common-init.mk
@@ -41,7 +41,8 @@ PRODUCT_PACKAGES += \
     qseecom.rc \
     rmt_storage.rc \
     sct_service.rc \
-    sensorspd.rc \
+    adsp-sensorspdr.rc \
+    sdsp-sensorspdr.rc \
     sensors.rc \
     sscrpcd.rc \
     tad.rc \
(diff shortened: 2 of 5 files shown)


MarijnS95:
Note that you're starting both services in different classes, and that `sdsp-sensorspdr` is not `disabled` and will thus start as soon as `core` starts, instead of waiting for `slpiup=1`.

Felix and me are planning ahead to move as much into BP as possible and have as little inexistant packages as possible. Part of that depends on defining variables earlier (in `lower-case.mk` makefiles instead of `UpperCase.mk`) which @haxk20 did for us in platform repositories (see his PRs). We'd move these files into a `prebuilt_etc` node in `rootdir/Android.bp` and have the `ifeq` in `common-init.mk`.

Finally, instead of that note in `sensors.rc` I'm personally fine moving
```
on property:vendor.qcom.slpiup=1
    enable vendor.sensorspd
```
Into the respective adsp and sdsp files. When checking for transitive `on property` dependencies all files are usually grepped for anyway.

kholk:
The missed "disabled" of course wasn't a wanted thing.... Btw, I wanted to do it like that as well, I'm only worried that sensors starts before sensorspd: in that case it will "work" but will not find any sensors.....

kholk:
Forcepushed to add the missing disable

rootdir/Android.mk
```diff
@@ -40,11 +40,26 @@ LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/etc/init
 include $(BUILD_PREBUILT)
 endif
 
-ifeq ($(TARGET_NEEDS_SENSORS_PDR), true)
+# Note: Sensors PDR can be on ADSP **or** on SDSP
+ifeq ($(TARGET_NEEDS_ADSP_SENSORS_PDR), true)
+ # Let's error out if the developer makes an impossible choice
+ ifeq($(TARGET_NEEDS_SDSP_SENSORS_PDR), true)
```
MarijnS95, line 46: `ifeq(` needs a space otherwise the parser throws up and (erroneously) fails on the following line:

    device/sony/common/rootdir/Android.mk:58: error: extraneous `else'.

MarijnS95:
> I'm only worried that sensors starts before sensorspd: in that case it will "work" but will not find any sensors.....

You should anyway: `enable` (and `start` too, for that matter) is asynchronous and allows the service to be started when the `class` it belongs to is set to be running. There is no guarantee that the service is immediately started while init waits for "some magic event to happen" before returning and executing the next line in init (in this case `sensors` after `sensorspd`).

In any case I assume both services can be started in parallel and are reasonably resilient to missing the other, for some time. In particular `sensors` retries the SSC connection for about 20s, which should be well enough for `sensorspd` to run, even if my change suggested above may start them in the wrong order.

If there's really a hard dependency we should introduce another property - or rather the binary behind `vendor.sensorspd` should set this for us.

In any case this IMMEDIATELY fixes the sensor issues on 55.1 - thanks! Unfortunately both `sscrpcd` invocations like to fail a ton of times and restart far too often: `sscrpcd daemon will restart after 25ms...`. Can we decrease this to 200ms or something?

rootdir/Android.mk
```diff
@@ -40,11 +40,26 @@ LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/etc/init
...
+
+include $(CLEAR_VARS)
+LOCAL_MODULE := adsp-sensorspdr.rc
+LOCAL_MODULE_CLASS := ETC
+LOCAL_SRC_FILES := vendor/etc/init/adsp-sensorspdr.rc
+LOCAL_MODULE_TAGS := optional
+LOCAL_MODULE_PATH := $(TARGET_OUT_VENDOR)/etc/init
+include $(BUILD_PREBUILT)
+
+else ifeq ($(TARGET_NEEDS_SDSP_SENSORS_PDR), true)
```
stefanhh0, line 58: FAILED: 
device/sony/common/rootdir/Android.mk:58: error: extraneous `else'.

MarijnS95:
As it stands this only works (is a requirement) on Kumano 55.1 firmware (Android 10), but fails and restarts every 25ms on 55.0 (Android 9) firmware.

The same happens on an Akatsuki with older Android 9 firmware - but `sscrpcd sensorspd` is started on Android 10 stock as well which likely suffers the same requirement as Kumano.

kholk:
Typo fixed and DNM status removed.

Please merge this PR.

PR #770, opened 2020-07-03
[COMMON] [Q-MR1] Add IMS/CnE 

kholk:
This patchset adds the IMS/CnE functionality on AOSP test-keys builds by
using the files that have been provided in the current ODM release.
Also cares about custom ROM developers, which are able to EXCLUDE the
test-keys signed apps from the auto bind mount, so that we don't force
compromising security of their system, leaving them the choice.


MartinX3:
Tested on Akari DSDS on German O2 (with VoLTE/VoWIFI support)

```
IMS-Service Status
IMS-Registration: Available
VoLTE: Available
VoWLAN: Not available
Video Calls: Not available
UT-UI: Available
```

After I forced LTE only:

```
IMS-Service Status
IMS-Registration: Not available
VoLTE: Not available
VoWLAN: Not available
Video Calls: Not available
UT-UI: Available
```
Then a call test just ended immediately without ringing.

Log:
[logcat-b_radio.log](https://github.com/sonyxperiadev/device-sony-common/files/4870248/logcat-b_radio.log)

kholk:
Was this tested on custom ROM or pure aosp?

Paulbouchara:
This is probably LOS

MartinX3:
On LineageOS.
Before the move from PE Custom to SODP I got even with forced LTE the status `Available` with IMS-Registration and VoLTE.
But there also the call stopped instantly without ringing.

On LOS I use the SODP repos and removed the conflicting LOS CAF ones.

kholk:
I need a urgent test of this PR on pure AOSP for a lot of reasons.

In any case, thanks for testing the custom rom usecase, I am happier when things are tested on both sides!
Regarding that symptom, luckily it's not breaking calls, as the VoLTE option can be disabled on your device's settings.

In any case, this may happen for many reasons, like authentication failure, bad setup for your carrier (CarrierConfig overlay!) or.. Well, anything really.

When I tried VoLTE calling with a TIM IT simcard it was working great, but my setup was slightly different... So, well, let's say...

This is the very very necessary part to get the functionality up: let's get this one done, then we think about the rest step by step!

kholk:
Merge

PR #778, opened 2020-09-12
[R/Q] android.bp: Remove hidltransport and hwbinder from shared libs

Haxk20:
In Android R libhwbinder and libhidltransport are no longer available to use.
Thus let's remove the libraries.

### DIFF
--- a/CommonConfig.mk
+++ b/CommonConfig.mk
@@ -27,6 +27,10 @@ TARGET_NO_BOOTLOADER := true
 TARGET_NO_RECOVERY ?= false
 TARGET_NO_KERNEL := false
 
+# Android R: Disable logic for new vendor_boot
+# Our devices do not support it
+TARGET_NO_VENDOR_BOOT := true
+
 # common cmdline parameters
 ifneq ($(BOARD_USE_ENFORCING_SELINUX),true)
 BOARD_KERNEL_CMDLINE += androidboot.selinux=permissive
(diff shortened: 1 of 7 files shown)


MarijnS95:
I would have appreciated a more elaborate explanation in the commit: why these libraries are still there but can't be linked from a `vendor` context, why there is no replacement -> whether this has always been a transitive (hence unnecessary) dependency, and finally if other basic dependencies (that are not directly referenced in code afaik) are transitively included as well such as `libhidlbase`.

Haxk20:
@MarijnS95 We have to remove them because Android 10 deprecates those 2 libraries. That doesnt mean that the stuff they provided no longer exists tho.
libhidlbase got the role of having all the symbols of libhidltransport and libhwbinder.

MarijnS95:
@Haxk20 Correct, that's the kind of explanation I expected in the commit, together with sources backing up these claims:

https://android.googlesource.com/platform/system/libhidl/+/8f65ba713e73b40dc58dd7a8a702a96d6c1c2181
https://android.googlesource.com/platform/system/libhidl/+/a46371d5b3ffd08808ae93ec420721345738ec65
https://android.googlesource.com/platform/system/libhwbinder/+/09a8725d56b168668a11530537c9738f4bc58a90

All three are on `android-10` tags as well which is why I asked you test and PR these for Q (and have R rebased on top of that).

jerpelea:
please push to Q-mr1

jerpelea:
cherry-picked

PR #781, opened 2020-09-15
[R]rootdir: init: Remove useless verity_load_state

sjllls:
On R, verity state is better to be checked by bootloader instead of init in userspace.
`do_verity_load_state` is not be supported on R.

https://android-review.googlesource.com/c/platform/system/core/+/946923/3/init/builtins.cpp#b710

Error:
`host_init_verifier: device/sony/common/rootdir/vendor/etc/init/hw/init.common.rc: 69: Invalid keyword 'verity_load_state'`


MarijnS95 (approved):
Neat, thanks for submitting this together with source!

@jerpelea Merge!

PR #720, opened 2020-05-01
[DNM] vintf: Align Secure Element

ix5:
### Add Secure Element interfaces

Guarded by `TARGET_HAS_SECURE_ELEMENT`. This will have to be set in the platform repos for devices which provide the secure element.

PR #792, opened 2020-10-05
[DONOTMERGE] [1/2] Tracking PR for secure element feature

MarijnS95:
This is a fixup of #720. As it turns out the PR might be slightly more complex to enable access to the SE in the NFC chip. Access to the SE in a UICC should be fine.

TODO:
- [ ] Seine and Kumano are the only devices with a PN557 from NXP which seems to have an "NFC secure element" but lack a config file (on stock and SODP); the HAL will crash without this. A variant of the file is easily available in the repo: https://cs.android.com/android/_/android/platform/hardware/nxp/secure_element/+/master:libese-spi/p73/libese-nxp-P73.conf but lack of this on stock makes me curious: how's that working?
- [ ] `SIM1` and `SIM2` are hosted by `qcrild`; do all our devices have access to the SE in the UICC, or is that limited for Seine and Kumano?
- [ ] @ix5 What's your feeling towards file naming? `vendor.nxp.nxpese.xml` should IMO get a name that represents the service from `hardware/nxp/secure_element` that hosts both `INxpEse` as well as `ISecureElement/eSE1`. Likewise `ISecureElement/SIM[12]` should be moved to an existing (renamed) SS/DS file representing `qcrild`, which hosts those.
     In other words, I think we should name vintf files (logically) after the services hosting them rather than the name of the service itself; that can be read from the contents. Unless you disagree, then it's best to move `eSE1` into `android.hardware.secure_element.xml` (wihout ss/ds postfix, of course).
- [ ] #720 was initially filed for Q, this should be backwards compatible and merged there.

### DIFF
--- a/CommonConfig.mk
+++ b/CommonConfig.mk
@@ -165,11 +165,26 @@ endif
 # SELinux
 include device/sony/sepolicy/sepolicy.mk
 
+# Device manifest: What HALs the device provides
 DEVICE_MANIFEST_FILE += $(COMMON_PATH)/vintf/manifest.xml
+# Framework compatibility matrix: What the device(=vendor) expects of the framework(=system)
 DEVICE_MATRIX_FILE   += $(COMMON_PATH)/vintf/compatibility_matrix.xml
 
-# Custom NXP vendor interfaces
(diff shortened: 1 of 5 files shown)


MarijnS95:
Closing until this is split out in a `SIM[12]` part for `qcrild` and `eSE1` for NFC SE reader.

PR #803, opened 2020-11-30
Adding several NOBA modem configurations as a fix #659 

AngryGami:
for https://github.com/sonyxperiadev/bug_tracker/issues/659

As suggested by @MarijnS95


Paulbouchara:
This seems good, more configs is always good, merge !

PR #804, opened 2020-12-23
[Master] common : Add a guard to allow use of other APNs config

Paulbouchara:
This is defaulted to yes so regular SODP builds won't be affected, but ROMs like PE enforce the use of their own APNs and that causes a double definition.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -133,9 +133,6 @@ PRODUCT_PACKAGES += \
 PRODUCT_PACKAGES += \
     firmware_folders
 
-# Community APN list
-PRODUCT_COPY_FILES += \
-    $(COMMON_PATH)/rootdir/vendor/etc/apns-conf.xml:system/etc/apns-conf.xml
 
 ifeq ($(MODEM_CONFIG_LEGACY_PLATFORM),true)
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -133,9 +133,13 @@ PRODUCT_PACKAGES += \
 PRODUCT_PACKAGES += \
     firmware_folders
 
+
+USE_SODP_APNS ?= true
+ifeq ($(USE_SODP_APNS),true)
 # Community APN list
 PRODUCT_COPY_FILES += \
     $(COMMON_PATH)/rootdir/vendor/etc/apns-conf.xml:system/etc/apns-conf.xml
+endif #USE_SONY_APNS
```
ix5, line 142: `USE_SODP_APNS` != `USE_SONY_APNS`

Paulbouchara: Thanks, force pushed

common.mk
```diff
@@ -133,9 +133,8 @@ PRODUCT_PACKAGES += \
 PRODUCT_PACKAGES += \
     firmware_folders
 
-# Community APN list
-PRODUCT_COPY_FILES += \
-    $(COMMON_PATH)/rootdir/vendor/etc/apns-conf.xml:system/etc/apns-conf.xml
+
+USE_SODP_APNS ?= true
```
MarijnS95, line 137: Why is `USE_SODP_APNS ?= true` here and was `PRODUCT_COPY_FILES += \ ... ` moved below?

Just move them all below `-include device/sony/customization/customization.mk` for consistency, `?=` will only set it to a value if it wasn't assigned to before.

PR #805, opened 2020-12-23
[q-mr1] odm: Add build.prop cleaner

ix5:
Certain props need to be filtered out from the proprietary `odm` image ("Software Binaries") because they are device-specific, while the `odm` image is released per platform.

### DIFF
--- a/common-odm.mk
+++ b/common-odm.mk
@@ -11,6 +11,10 @@ BOARD_ODMIMAGE_FILE_SYSTEM_TYPE := ext4
 PRODUCT_ODM_PROPERTIES += \
     ro.odm.version=$(PLATFORM_VERSION)_$(SOMC_KERNEL_VERSION)_$(SOMC_PLATFORM)_$(TARGET_VENDOR_VERSION)
 
+# Clear device-specific info from platform-level odm image
+PRODUCT_PACKAGES += \
+    odm_prop_cleaner
+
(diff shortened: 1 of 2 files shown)


MarijnS95:
Can't shouldn't this somehow depend on the file being created first? Otherwise this build rule _might_ run before it is created?

Also, any idea why it ends up in `/odm/etc/build.prop` instead of `/odm/build.prop`?

ix5:
> Can't shouldn't this somehow depend on the file being created first? Otherwise this build rule _might_ run before it is created?

You could use `LOCAL_REQUIRED_MODULES`, but afaics that does not guarantee execution ordering, rather only adds the `required` module to `PRODUCT_PACKAGES`.

If you've got a good idea, shoot.

> Also, any idea why it ends up in `/odm/etc/build.prop` instead of `/odm/build.prop`?

[El Goog CATD](https://android-review.googlesource.com/c/platform/build/+/964771).

jerpelea:
you can just create a file /odm/build.prop

ix5:
> you can just create a file /odm/build.prop

`/odm/build.prop`  will not be recognized by the property service on Q anymore https://android.googlesource.com/platform/system/core/+/refs/tags/android-10.0.0_r41/init/property_service.cpp#892

Of course we could just manually read the file, but I think for now this is a good solution.

PR #808, opened 2021-01-26
hardware/qcom/Android.mk: Add sm8250 (KONA) to media platform lists.

Haxk20:
Without this we would fallback to sdm660-libion media HAL which is wrong.
Add sm8250 to the lists so that the correct HALs are used.


MarijnS95:
@Haxk20 It looks like our convention was broken since Seine, think we should use `$(TRINKET)` and `$(KONA)` here.

MarijnS95:
Whoops, should have submitted that earlier - it has already been merged.

PR #810, opened 2021-02-06
[master] hardware: ssc: Kona sensors

ix5:
**DONOTMERGE** Just for bringup purposes
@kholk

### DIFF
--- a/hardware/ssc/Android.bp
+++ b/hardware/ssc/Android.bp
@@ -472,3 +472,243 @@ prebuilt_etc {
     sub_dir: "sensors/config",
     vendor: true,
 }
+prebuilt_etc {
+    name: "kona_ak991x_0.json",
+    src: "config/kona_ak991x_0.json",
+    sub_dir: "sensors/config",
+    vendor: true,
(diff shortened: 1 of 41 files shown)


kholk:
The first commit is a sure merge: please, split them and let's get the first one in asap.

ix5:
Pushed off the updated non-edo-specific configs.

Someone please build-test this!

kholk:
I think @MarijnS95 has tested it already, since he gave me an image with these files in, but I'm not sure whether it was your commit or a self-baked one.

Please Marijn, can you confirm?

MarijnS95:
Apologies, I have not added these files yet - so they shouldn't be in your image either @kholk! Well, that is, until now, the patch is now in my tree and will be in the next vendor image :grimacing:

MarijnS95 (approved):
Edo is magical. This is the first platform where sensors "just work" out of the box, no random crashes around SSC because we're missing an obscure `*pd` userspace binary somewhere, or corruption on persist.

Thanks!

PR #815, opened 2021-02-06
[COMMON] init: Add pd_mapper service to SM8250 platforms

kholk:
The SM8250 platforms have PDRs discoverable with pd_mapper.


kholk:
.....but then Marijn also did tftp_server, which is also needed, so I'm closing mine, merge his one

MarijnS95:
> but then Marijn also did tftp_server

You also did, they're in the same `ifneq` block ;)

PR #816, opened 2021-02-06
[COMMON] [MASTER] rootdir: Unconditionally provide pd_mapper and tftp_server rcs

MarijnS95:
All boards listed here and new (unlisted!) SM8250 are the only ones supported by SODP R. Remove the guard entirely.


kholk:
FIXME LATER: throw this to Android.bp, since guarding is not necessary anymore

MarijnS95:
Already moved to Android.bp :)

PR #819, opened 2021-02-08
[COMMON] [R] Migrate to display stack at CAF 9.1[25] branches/tags

MarijnS95:
Notable changes include:

- Finally! Vintf fragments! Now we don't need to keep track of these ourselves anymore, the module itself automatically supplies them to the build system;
- Composer is now its own executable HAL, instead of a libhardware module loaded by a passthrough (AOSP) service;
- Blobs link against QTI `mapper@2.0`; EGL fails to probe without it.

We _cannot_ merge this before knowing the impact on other platforms! As usual the HAL _might_ run on 4.14 devices, but more than likely it does not. Tough luck, we'll need to come up with guards around all these changes.


kholk:
No you cannot get the new HALs on the 4.14 BSPs, you will get incompatibilities with the proprietary sdmextension.

Add the proper barriers for the 4.19 based projects and #shipit

PR #822, opened 2021-02-12
[COMMON] [master] packages: Drop inexistant libloc_pla and libloc_ds_api

MarijnS95:
Not applying this to Q, older branches (A10 r1) still reference the sdm845 HAL and some other branch where these _may_ still exist. Even then these should likely all be transitive dependencies that are dynamically linked.

---

These GPS libraries do not exist on 8.1 nor 9.12 branches. We do not have any enforcement for existing packages in `PRODUCT_PACKAGES` just yet, but when we end up enabling that these will pop out. They are probably reminiscent of some ancient HAL 3 Android versions ago.

### DIFF
--- a/common-odm.mk
+++ b/common-odm.mk
@@ -648,6 +648,7 @@ PRODUCT_PACKAGES += \
     com.qti.node.swregistration \
     com.qti.camx.chiiqutils \
     libcom.qti.chinodeutils \
+    libcamera_nn_stub \
     com.qti.chi.override \
     com.qti.node.eisv2 \
     com.qti.node.eisv3 \


ix5:
Drop `libqsocket` as well while you're at it https://github.com/sonyxperiadev/vendor-qcom-opensource-location/pull/31#discussion_r575657443

MarijnS95:
That's an ODM dep though: 2ecc882b4c66cb1496f08248ff6605d30e7e340a

ix5:
Roger

MarijnS95:
@ix5 Shall we revert that commit anyway? Perhaps up to @jerpelea to take up with his proprietaries :)

ix5:
Revert away, I say

MarijnS95:
Doesn't seem like such a good idea in the end. That commit mentions `tftpd?_server`:

```console
$ readelf -d mnt/bin/tftp_server

Dynamic section at offset 0x191f8 contains 33 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libqsocket.so]
...
```

ix5:
Oh I thought you were talking about the commit in location OSS. No, ofc alin can't drop libqsocket then.

MarijnS95:
@ix5 Ah, right. It looks like @kholk was too late to drop them, PR [already merged](https://github.com/sonyxperiadev/vendor-qcom-opensource-location/pull/31#event-4330188851). Oh well, we'll drop the extra baggage another time.

jerpelea:
they can be dropped in the next PR

MarijnS95:
@jerpelea Nah, I don't want explicit reverts.

I think this is ready for merging then, though :wink

PR #823, opened 2021-02-13
[COMMON] vendor: init: sscrpcd: Add wakelock group to sensor dsp daemon

kholk:
This daemon can set a wakelock and will fail to initialize if it
has no permission to.


MarijnS95 (approved):
Nice, it stopped crashing and restarting every 25ms.

Merge!

PR #826, opened 2021-02-18
[r-mr1] rootdir: rmt_storage: Enabled based on baseband

ix5:
Fixes confusing logspam for `mdm` targets:
```
E vendor.rmt_storage: Remote storage service is not supported on mdm target
```


MarijnS95:
Should have probably mentioned that after printing this `rmt_storage` tries to disble itself by setting `ctl.stop` to `vendor.rmt_storage` - which is obviously denied by sepolicy.

PR #827, opened 2021-02-18
[common] [R] hardware/camera: Emit symlinks for sm8250

MarijnS95:
Sorry, forgot to submit this yesterday!

---

CamX on Edo needs these symlinks. We should probably come up with a better method of representing this instead of hardcoding SoC names in common.


MarijnS95:
Oh well. I intended to force-push this to `$(KONA)` but we don't use that anywhere else in common either...

PR #828, opened 2021-02-18
[COMMON] Add new services for 4.19-based BSPs

kholk:
This patchset adds the new services for 4.19 BSPs and Kona.

### DIFF
--- a/common-init.mk
+++ b/common-init.mk
@@ -68,7 +68,8 @@ PRODUCT_PACKAGES += \
 # Secure Processor
 ifeq ($(TARGET_HAS_SPU),true)
 PRODUCT_PACKAGES += \
-    spu.rc    
+    authsecret.rc \
+    spu.rc
 endif
 
 # modemswitcher
--- a/common-init.mk
+++ b/common-init.mk
@@ -65,6 +65,12 @@ PRODUCT_PACKAGES += \
     init.qcom.ipastart.sh \
     init.qcom.slpistart.sh
 
+# Secure Processor
+ifeq ($(TARGET_HAS_SPU),true)
+PRODUCT_PACKAGES += \
+    spu.rc    
+endif
+
 # modemswitcher
 PRODUCT_PACKAGES += \
     vendor.somc.hardware.miscta@1.0-service.rc \
--- a/rootdir/vendor/etc/init/spu.rc
+++ b/rootdir/vendor/etc/init/spu.rc
@@ -1,0 +1,7 @@
+service spu_service /odm/bin/hw/vendor.qti.spu@1.0-service
+    class hal
+    user system
+    group system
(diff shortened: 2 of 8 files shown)


rootdir/vendor/etc/init/spu.rc
```diff
@@ -0,0 +1,8 @@
+service spu_service /odm/bin/hw/vendor.qti.spu@1.0-service
+    class hal
+    user system
+    group system
+    disabled
+
+on boot
+    start spu_service
```
MarijnS95, line 8: Eww, can't they put it in the right class... Or at least `enable`!!! it instead of `start`ing so that it restarts when it _inevitably_ crashes.

MarijnS95: Please just sure `class core` and remove this + `disabled` BS.

https://sx.ix5.org/info/android-init-stages/

kholk: It's like that on purpose... this service is essential for security... that's why....

common-init.mk
```diff
@@ -65,6 +65,13 @@ PRODUCT_PACKAGES += \
     init.qcom.ipastart.sh \
     init.qcom.slpistart.sh
 
+# Secure Processor
+ifeq ($(TARGET_HAS_SPU),true)
+PRODUCT_PACKAGES += \
+    authsecret.rc
+    spu.rc
```
MarijnS95, line 72:
```suggestion
PRODUCT_PACKAGES += \
    authsecret.rc \
    spu.rc    
```

```
device/sony/common/common-init.mk:72: error: missing separator.
```

MarijnS95: By the way you haven't declared prebuilt packages for any of these.

kholk:
Branch forcepushed: added the requested fixes.

rootdir/vendor/etc/init/spu.rc
ix5: If you removed the `disabled`, there's no need to keep the `on boot` trigger since `class_start hal`/`class_start_post_data` will occur before the `on boot` target is reached.

MarijnS95: @ix5 the stock rc file does it like this, `class hal` with `on boot start`. But apparently `class hal` starts earlier than `boot` making this useless?

ix5: Stock idiocy is not the benchmark ;)

MarijnS95: > Stock idiocy is not the benchmark ;)

It isn't our intent to match with stock; rather this file seems to be copied from stock but the `disabled` appeared out of thin air...

PR #834, opened 2021-03-10
[r-mr1] [3/3] common.mk: common-headers soong namespaces

ix5:
Add a soong namespace for kernel/sony/msm-X.Y/common-headers based on platform kernel version so that different kernels can live side-by-side

See:
https://github.com/sonyxperiadev/device-sony-common-headers/pull/12
https://github.com/sonyxperiadev/device-sony-common-headers/pull/13


ix5:
Let's find a better way

PR #837, opened 2021-03-18
[DNM][S] common.mk: Remove PRODUCT_BUILD_BOOT_IMAGE

Haxk20:
As of Android S enabling PRODUCT_BUILD_BOOT_IMAGE and
BOARD_USES_RECOVERY_AS_BOOT is not allowed:

    PRODUCT_BUILD_BOOT_IMAGE is true, but so is BOARD_USES_RECOVERY_AS_BOOT.
    Skipping building boot image.

Some of the devices we support enable BOARD_USES_RECOVERY_AS_BOOT.
Remove PRODUCT_BUILD_BOOT_IMAGE to build boot image on those devices.


Haxk20:
Yes every device that needs the removed definition will have to define it in device repo as google made both definitions here only settable once.

PR #838, opened 2021-03-29
[DNM] [r-mr1] hardware: Add qti-headers intermediary module

ix5:
Selects `qti_kernel_headers` repo based on kernel version so that multiple versions can live inside on AOSP tree.

Needs following changes:

```make
SOONG_CONFIG_NAMESPACES += qti_kernel_headers
SOONG_CONFIG_qti_kernel_headers := version
SOONG_CONFIG_qti_kernel_headers_version := 4.14
```

Inisde `common-headers` needs a rename of module name from `qti_kernel_headers` to `qti_kernel_headers_4.14` (based on
version).

This will finally allow multiple kernel versions (and BSPs) to live side-by-side again.

### DIFF
--- a/hardware/qti-headers/Android.bp
+++ b/hardware/qti-headers/Android.bp
@@ -1,0 +1,32 @@
+// Select kernel headers based on version
+
+// Use as follows:
+// SOONG_CONFIG_NAMESPACES += qti_kernel_headers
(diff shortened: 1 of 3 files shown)


ix5:
This isn't going to work in device-sony-common because of the root-level Android.bp declaring a soong namespace.
Needs its own repo

PR #840, opened 2021-03-31
[r-mr1] common: Set qti-headers kernel version via soong

ix5:
This change will allow multiple kernel versions to live side-by-side in one AOSP tree.

Needs to be accompanied by changes to kernel-headers repos to change module name to `qti_kernel_headers_X.Y` and an
auxiliary module to resolve `qti_kernel_headers` to the correct module.

Overview:
- https://github.com/sonyxperiadev/device-sony-common/pull/839
- https://github.com/sonyxperiadev/device-sony-common-headers/pull/16
- https://github.com/sonyxperiadev/device-sony-common-headers/pull/15
- https://github.com/ix5/qti-headers (to be forked and then placed into `kernel/sony/qti-headers`)

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -70,6 +70,12 @@ KERNEL_PATH := kernel/sony/msm-$(SOMC_KERNEL_VERSION)
 # Sanitized prebuilt kernel headers
 -include $(KERNEL_PATH)/common-headers/KernelHeaders.mk
 
+# Configure qti-headers auxiliary module via soong so that the correct headers
+# under kernel/sony/msm-X.Y/kernel-headers are chosen
+SOONG_CONFIG_NAMESPACES += qti_kernel_headers
+SOONG_CONFIG_qti_kernel_headers := version
+SOONG_CONFIG_qti_kernel_headers_version := $(SOMC_KERNEL_VERSION)
+
 # Codecs Configuration
 PRODUCT_COPY_FILES += \
     frameworks/av/media/libstagefright/data/media_codecs_google_audio.xml:$(TARGET_COPY_OUT_VENDOR)/etc/media_codecs_google_audio.xml \


MarijnS95 (approved):

common.mk
```diff
@@ -70,6 +70,12 @@ KERNEL_PATH := kernel/sony/msm-$(SOMC_KERNEL_VERSION)
 # Sanitized prebuilt kernel headers
 -include $(KERNEL_PATH)/common-headers/KernelHeaders.mk
 
+# Configure qti-headers auxiliary module via soong so that the correct headers
+# under kernel/sony/msm-X.Y/kernel-headers are chosen
+SOONG_CONFIG_NAMESPACES += qti_kernel_headers
```
MarijnS95, line 75: We got a nice `+=` here to not accidentally mess with what is set in `CommonConfig.mk` for `gralloc`... Fortunately that one uses `+=` too.

Which, by the way, why wasn't that moved from `CamelCase.mk` to product-inheritance `lowercase.mk`?

ix5: We can probably remove the gralloc thing entirely now, no?
I don't know of any A11 devices that set `TARGET_USES_GRALLOC1`.

MarijnS95:
```console
me:sodp/device/sony $ rg GRALLOC1
yoshino/platform.mk
74:TARGET_USES_GRALLOC1 := true

nile/platform.mk
74:TARGET_USES_GRALLOC1 := true

ganges/platform.mk
71:TARGET_USES_GRALLOC1 := true

loire/PlatformConfig.mk
72:TARGET_USES_GRALLOC1 := true
```

(Note there is no Tone checkout in my tree, neither is Loire relevant)

ix5: Whoops, had my trees checked on the wrong branches locally. Do you wanna submit the change to lowercase?

PR #847, opened 2021-05-20
[r-mr1] common: Fix ims bindmount path

sjllls:
IMS is broken with the wrong path, although it is still broken after this fix...

### DIFF
--- a/rootdir/vendor/etc/init/bindmount-apps.rc
+++ b/rootdir/vendor/etc/init/bindmount-apps.rc
@@ -1,7 +1,7 @@
 on post-fs
     # Bind mount all priv-apps
     mount none /odm/system_ext/priv-app/dpmserviceapp /system/system_ext/priv-app/dpmserviceapp bind rec
-    mount none /odm/system-ext/priv-app/ims /system/system_ext/priv-app/ims bind rec
+    mount none /odm/system_ext/priv-app/ims /system/system_ext/priv-app/ims bind rec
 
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
Nice catch!

Paulbouchara:
Oh my that typo....
Well yeah ims is broken and you need supplementary apps to make it work aka fetch from PE, attention if you're running enforcing you might have RIL crashing like a champ

sjllls:
> Oh my that typo....
> Well yeah ims is broken and you need supplementary apps to make it work aka fetch from PE, attention if you're running enforcing you might have RIL crashing like a champ

Thanks, I will try it, why did IMS doesn't work with these apps in odm parition and PE choose to include them in device_sony_customization?

sjllls:
@jerpelea q-mr1 branch also needs this commits, thanks!

MarijnS95:
> @jerpelea q-mr1 branch also needs this commit, thanks!

Oopsie, should have submitted to `q-mr1` so that it can be rebased into R :) - for next time :)

MarijnS95:
> Thanks, I will try it, why did IMS doesn't work with these apps in odm parition and PE choose to include them in device_sony_customization?

We played around trying to make Android accept the signature of the APKs and give them extra permissions. That's easily doable, but their signature must be **the same as those in the radio group** because of `sharedUserId` which depends on local build keys, hence the need to re-sign the apks with those local keys as @Paulbouchara mentioned.

PR #850, opened 2021-05-21
[q-mr1] Fix building

sjllls:
Display hal moved into sm8150 dir.

Soong namespace path changed after https://github.com/sonyxperiadev/hardware-qcom-display/commit/fbfe48f6547007449ac7431f20733b79b9c171c3

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -12,15 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Common path
+COMMON_PATH := device/sony/common
+
+display_platform := sm8150
+
 # Enable building packages from device namspaces.
 # Might be temporary! See:
 # https://android.googlesource.com/platform/build/soong/+/master/README.md#name-resolution
 PRODUCT_SOONG_NAMESPACES += \
     $(PLATFORM_COMMON_PATH) \
-    vendor/qcom/opensource/core-utils
-
-# Common path
-COMMON_PATH := device/sony/common
+    vendor/qcom/opensource/core-utils \
+    vendor/qcom/opensource/display/$(display_platform) \
+    vendor/qcom/opensource/display-commonsys-intf/$(display_platform)
 
 # Build scripts
 SONY_CLEAR_VARS := $(COMMON_PATH)/sony_clear_vars.mk
--- a/hardware/qcom/Android.mk
+++ b/hardware/qcom/Android.mk
@@ -25,7 +25,7 @@ endif
 OMX_VIDEO_PATH := mm-video-v4l2
 
 SRC_CAMERA_HAL_DIR ?= vendor/qcom/opensource/camera
-SRC_DISPLAY_HAL_DIR := vendor/qcom/opensource/display
+SRC_DISPLAY_HAL_DIR := vendor/qcom/opensource/display/sm8150
 SRC_MEDIA_HAL_DIR := $(QCOM_MEDIA_ROOT)
 TARGET_KERNEL_VERSION := $(SOMC_KERNEL_VERSION)
 


CommonConfig.mk
```diff
@@ -14,6 +14,9 @@
 
 -include device/sony/customization/Customization.mk
 
+#Broken Rules
+BUILD_BROKEN_DUP_RULES := true
```
MarijnS95, line 18: That's very wrong, the error should be fixed instead.

MarijnS95:
> Display hal moved into sm8150 dir.

Only on Android 11 (R), Q is still at the normal location and will always be: https://github.com/sonyxperiadev/local_manifests/blob/b63f361a36c6d85aa2dd66feb08220648702c0ac/qcom.xml#L9-L11. Which `local_manifests` branch have you synced against?

CommonConfig.mk
sjllls: I didn't find the duplicate define of 'vendor.qti.hardware.display.composer-service.xml' in hardware or vendor. So still checking the reason.

MarijnS95:
https://github.com/sonyxperiadev/vendor-qcom-opensource-commonsys-intf-display/commits/aosp/LA.UM.8.1.r1 https://github.com/sonyxperiadev/hardware-qcom-display/commits/aosp/LA.UM.8.1.r1 - I think the last few commits are problematic for Q?

sjllls:
> https://github.com/sonyxperiadev/vendor-qcom-opensource-commonsys-intf-display/commits/aosp/LA.UM.8.1.r1 https://github.com/sonyxperiadev/hardware-qcom-display/commits/aosp/LA.UM.8.1.r1 - I think the last few commits are problematic for Q?

Yup. soong namespace changed. It point to a wrong dir.

MarijnS95:
https://github.com/sonyxperiadev/local_manifests/pull/126 That's the missing part of the puzzle :)

Do mention it as a "Depends on https://github.com/sonyxperiadev/local_manifests/pull/126" in the PR description, makes it easier to understand what's going on 😉

sjllls:
> [sonyxperiadev/local_manifests#126](https://github.com/sonyxperiadev/local_manifests/pull/126) That's the missing part of the puzzle :)
> 
> Do mention it as a "Depends on [sonyxperiadev/local_manifests#126](https://github.com/sonyxperiadev/local_manifests/pull/126)" in the PR description, makes it easier to understand what's going on 😉

Thanks!

common-treble.mk
```diff
@@ -12,8 +12,6 @@ PRODUCT_PACKAGES += \
     vendor.qti.hardware.display.mapper@2.0.vendor
 
 # Graphics allocator/mapper
-ifeq ($(TARGET_HARDWARE_GRAPHICS_V3),true)
```
MarijnS95, line 11: Is this true for Q too? Are Loire and Tone still somewhat alive/buildable here?

CommonConfig.mk
MarijnS95:
```
vendor/qcom/opensource/display/sm8150/sdm/libs/hwc2/Android.mk
65:LOCAL_VINTF_FRAGMENTS         := vendor.qti.hardware.display.composer-service.xml

vendor/qcom/opensource/display/sm8250/composer/Android.mk
73:    LOCAL_VINTF_FRAGMENTS         := vendor.qti.hardware.display.composer-service.xml
```

That's where they exist. Maybe they live in `hardware/` in some Q repo? Can you run this in the root of your AOSP tree:

```sh
grep vendor.qti.hardware.display.composer-service.xml device vendor hardware
```
And report the result?

common-treble.mk
sjllls: I am checking these commits agian, I think we do not need all of them.

MarijnS95:
By the way, do make a mention (for example `[sjllls: Remove handling for SM8250 which is not used on Q trees]` at the end of 881b4ab6b0cd8d1b452fb911846374095b25834b) in all the commits that were modified so that it is clear they are **not** the same as the commits on R :)

common.mk
```diff
@@ -12,15 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Common path
+COMMON_PATH := device/sony/common
```
MarijnS95, line 16: This one might have been a conflict resolution mistake? 008f9c214db1dc24a49587f9400f75c913da625e doesn't have it :)

sjllls: > This one might have been a conflict resolution mistake? [008f9c2](https://github.com/sonyxperiadev/device-sony-common/commit/008f9c214db1dc24a49587f9400f75c913da625e) doesn't have it :)

Yup, this is the result after solving cherry-pick conflict. For Q we only need to define sm8150.

sjllls (commented):
> By the way, do make a mention (for example `[sjllls: Remove handling for SM8250 which is not used on Q trees]` at the end of [881b4ab](https://github.com/sonyxperiadev/device-sony-common/commit/881b4ab6b0cd8d1b452fb911846374095b25834b)) in all the commits that were modified so that it is clear they are **not** the same as the commits on R :)

Okk, so I need to change the commit info.

CommonConfig.mk
sjllls:
```
sjll@buildhost:~/aosp$ grep vendor.qti.hardware.display.composer-service.xml -r device vendor hardware
Binary file vendor/qcom/opensource/display/sm8150/.git/index matches
vendor/qcom/opensource/display/sm8150/sdm/libs/hwc2/Android.mk:LOCAL_VINTF_FRAGMENTS         := vendor.qti.hardware.display.composer-service.xml
sjll@buildhost:~/aosp$ 
```

MarijnS95: That's correct, I don't really understand why you were seeing this duplicate copy error.

Is it gone now after recent changes?

common.mk
MarijnS95: Yes, don't forget to remove it then :)

CommonConfig.mk
sjllls: > That's correct, I don't really understand why you were seeing this duplicate copy error.
> 
> Is it gone now after recent changes?

No idea, but this error exists with LOCAL_VINTF_FRAGMENTS. :(

MarijnS95: Is the makefile where these are defined somehow imported twice, or are you running a tree with more than one display HAL that might provide this fragment? I really don't want to make Q run the inverse of R with https://github.com/sonyxperiadev/hardware-qcom-display/pull/42 :/

sjllls: > Is the makefile where these are defined somehow imported twice, or are you running a tree with more than one display HAL that might provide this fragment? I really don't want to make Q run the inverse of R with [sonyxperiadev/hardware-qcom-display#42](https://github.com/sonyxperiadev/hardware-qcom-display/pull/42) :/

Build system added it twice.
```
FAILED: 
build/make/core/base_rules.mk:526: error: overriding commands for target `out/target/product/pdx201/vendor/etc/vintf/manifest/vendor.qti.hardware.display.composer-service.xml', previously defined at build/make/core/base_rules.mk:526
10:05:35 ckati failed with: exit status 1
```

https://android.googlesource.com/platform/build/+/refs/tags/android-10.0.0_r41/core/base_rules.mk#526

MarijnS95: Yeah, the Q build system is broken. We don't really even use vintf fragments in Q yet except in GPS, for this reason most likely (Q is ancient enough I forgot it wasn't working properly here :grimacing:).

Since the namespace has to be removed anyway to circumvent:

```
error: vendor/qcom/opensource/display-commonsys-intf/sm8150/Android.bp:5:1: "soong-gralloc_defaults" depends on undefined module "blueprint"
Module "soong-gralloc_defaults" is defined in namespace "vendor/qcom/opensource/display-commonsys-intf/sm8150" which can read these 0 namespaces: []
Module "blueprint" can be found in these namespaces: ["."]
```

Let's go for legacy display Q branches and do it the messy way - as inflicted by AOSP.

MarijnS95:
For the record: we're planning to see if we can migrate all R devices to an R-based branch (CAF 8.1 is for Q) and reset Q back to a working state, so that none of these PRs are necessary anymore. Still, thanks for your contribution @sjllls! The end result would have been almost perfect if AOSP wasn't broken!

sjllls:
> For the record: we're planning to see if we can migrate all R devices to an R-based branch (CAF 8.1 is for Q) and reset Q back to a working state, so that none of these PRs are necessary anymore. Still, thanks for your contribution @sjllls! The end result would have been almost perfect if AOSP wasn't broken!

No problem, life is good now. :)

PR #853, opened 2021-07-05
rootdir: apns: Germany: O2: Enable IPv6

MartinX3:
And sync with https://www.o2apn.de/


MarijnS95:
Quick, change the target branch to `r-mr1` before it's too late!

(And perhaps to `q-mr1` so that it's all rebased on top)

MartinX3:
Ahhhh, ninjaaaaaaa changed. Thank you @MarijnS95 !
Sadly I can't choose 2 branches (which is of course normal, since the code could be different in the other branch.

MarijnS95:
Yeah you can't choose `q-mr1` unless rebasing it on Q while removing all commits that are only on R but not on Q.

PR #859, opened 2021-08-29
[q-mr1] Fix USB composition selector

bartcubbins:
sys.usb.configfs from init.usb.rc is triggering generic rules found in
/system/etc/init/hw/init.usb.configfs.rc which leads to the fact that
the values that correspond to the real configuration of the device are
overwritten by the generic (dummy) ones. Because of this, some USB modes
such as USB tethering, cannot be enabled in the USB Preferences menu.
Hence, change the prefix of sys.usb.configfs setting from "sys." to
"vendor." in order to avoid triggering the behavior described above.

The change was successfully tested on Xperia 10+ (Mermaid) device,
MTP, PTP, MIDI and USB tethering modes are enabled successfully and each
of them was recognized by the PC.

### DIFF
--- a/rootdir/vendor/etc/init/init.usb.rc
+++ b/rootdir/vendor/etc/init/init.usb.rc
@@ -112,6 +112,9 @@ on property:sys.usb.config=mtp && property:sys.usb.configfs=1
     setprop sys.usb.state ${sys.usb.config}
 
 # USB Mode : adb
+on property:sys.usb.config=adb && property:sys.usb.configfs=1
+    start adbd
+
 on property:sys.usb.ffs.ready=1 && property:sys.usb.config=adb && property:sys.usb.configfs=1
     write /config/usb_gadget/g1/configs/b.1/strings/0x409/configuration "adb"
     rm /config/usb_gadget/g1/configs/b.1/f1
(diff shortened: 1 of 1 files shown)


MarijnS95:
> [master]

Isn't this applicable to R and perhaps Q as well? Would make for some nice cleanup.

bartcubbins:
yes, this applies to all branches

MarijnS95:
If you can rebase this on Q and target `q-mr1`, Alin's scripts will pick the task of rebasing R and master on top automagically.

bartcubbins:
done!

MarijnS95 (approved):
Haven't tested, and I trust the first patch to be moving around compositions _without_ modifications (those would make the patch impossible to read/review), but the idea and justification behind the second patch look good to me.

MarijnS95:
Looking a bit more into this, we're already using the `vendor.sys.` prefix for runtime things, though only for `vendor.sys.keymaster.loaded` and `vendor.sys.listeners.*`.

bartcubbins:
Yes, it seems to me that vendor.sys. prefix is more suitable, I will update it soon

bartcubbins:
And I forgot about one thing in the first path. I will split it

bartcubbins:
now all the commits and their descriptions match what they do. sorry, working on a lot of different code, sometimes things like this get lost from view

MarijnS95:
> sorry, working on a lot of different code, sometimes things like this get lost from view

No worries, I'm actually quite glad someone serious is finally working on SODP, and that you're cleaning up commits on your own regard without needing to be asked :grin: 

> [1] https://github.com/FrankenRomP/android_system_core/blob/pie/adb/daemon/usb.cpp#L300

Can we link to official sources instead? This has moved: https://cs.android.com/android/platform/superproject/+/android-11.0.0_r40:system/core/adb/daemon/usb_ffs.cpp;l=301-302

bartcubbins:
oh, I just really love (get used) to use the github search and forgot to replace the link with the official one. updated.

MarijnS95:
@bartcubbins Github search is fine as long as you end up on `aosp-mirror` :stuck_out_tongue:, and reference a relatable tag/commit :)

PR #861, opened 2021-09-09
vintf: Add a build variable and manifest for NXP ESE

konradybcio:


Haxk20 (approved):

bartcubbins:
maybe it's better to place it next to the NXP NFC fragment?
https://github.com/sonyxperiadev/device-sony-common/blob/master/CommonConfig.mk#L180

jerpelea:
not all devices have SE
please add it to the platform

PR #862, opened 2021-09-15
[Q] hardware/adreno: Convert vendor->odm symlink to ro.hardware property

MarijnS95:
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/718

Android is continuously clamping down on symlink hacks (see the report in [1]) and for a good reason: they provide a perfectly valid alternative using `ro.hardware.<module name>` to change the platform name of a module that needs to be searched for.  On all our platforms the Vulkan library is called `vulkan.qcom.so` instead of `vulkan.$(TARGET_BOARD_PLATFORM).so` and Android can simply be taught to search for the former instead of the latter by setting `ro.hardware.vulkan=qcom`.

[1]: https://github.com/sonyxperiadev/bug_tracker/issues/718

Signed-off-by: Marijn Suijten <marijn.suijten@somainline.org>

### DIFF
--- a/common-prop.mk
+++ b/common-prop.mk
@@ -303,3 +303,7 @@ endif
 # Reduce cost of scrypt for FBE CE decryption
 PRODUCT_PROPERTY_OVERRIDES += \
     ro.crypto.scrypt_params=15:3:1
+
+# Look for vulkan.qcom.so instead of vulkan.$(BOARD_TARGET_PLATFORM).so
+PRODUCT_PROPERTY_OVERRIDES += \
+    ro.hardware.vulkan=qcom
--- a/hardware/adreno/Android.mk
+++ b/hardware/adreno/Android.mk
@@ -18,19 +18,13 @@ library_names := \
     libllvm-glnext.so \
     libllvm-qcom.so \
     librs_adreno.so \
-    librs_adreno_sha1.so \
-    hw/vulkan.qcom.so
+    librs_adreno_sha1.so
 
 # Create symlinks to 32- and 64-bit directories:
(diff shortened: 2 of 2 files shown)


bartcubbins:
Vulkan is still broken without the symlink. Just tested on Nile.

MarijnS95:
@bartcubbins any information at all that might help understand what is going on? Logs?

bartcubbins:
`09-20 15:10:09.397  4570  4570 E vndksupport: Could not load vulkan.qcom.so from sphal namespace: dlopen failed: library "vulkan.qcom.so" not found.`

https://android.googlesource.com/platform/frameworks/native/+/refs/tags/android-11.0.0_r45/vulkan/libvulkan/driver.cpp#182

MarijnS95:
@bartcubbins Correct, it loads from an SPHAL, and `/odm/lib64/hw/` is **not** included in the search path. `/odm/lib64/` is however: 
https://cs.android.com/android/platform/superproject/+/master:system/linkerconfig/contents/namespace/sphal.cc;l=40-47?q=sphal.cc. Making a symlink:

```console
griffin:/odm/lib64 # ls -l /odm/lib64/vulkan.qcom.so
lrwxrwxrwx 1 root root 17 2021-09-28 21:20 /odm/lib64/vulkan.qcom.so -> hw/vulkan.qcom.so
```

No matter how dirty, works.

MarijnS95:
Restoring part of the old behaviour in https://github.com/sonyxperiadev/device-sony-common/pull/865, validated on Griffin.

PR #871, opened 2021-11-04
[TEMP] common.mk: Disable compressed APEX

Haxk20:
As a temp fix lets disable compressed APEX which causes a12 to not boot on
4.14 kernel.

Please test !!!!!

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -68,6 +68,11 @@ PRODUCT_PACKAGES += \
 # Force building a recovery image: Needed for OTA packaging to work since Q
 PRODUCT_BUILD_RECOVERY_IMAGE := true
 
+ifeq ($(SOMC_KERNEL_VERSION), 4.14)
+# TEMP Disable compressed APEX
+PRODUCT_COMPRESSED_APEX := false
+endif
+
 # Android R: Disable logic for new vendor_boot
 # Our devices do not support it
 PRODUCT_BUILD_VENDOR_BOOT_IMAGE := false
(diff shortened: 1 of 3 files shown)


common.mk
```diff
@@ -68,6 +68,11 @@ PRODUCT_PACKAGES += \
 # Force building a recovery image: Needed for OTA packaging to work since Q
 PRODUCT_BUILD_RECOVERY_IMAGE := true
 
+ifneq ($(filter 4.14, $(SOMC_KERNEL_VERSION)),)
```
MarijnS95, line 71:
```suggestion
ifeq ($(SOMC_KERNEL_VERSION), 4.14)
```

Work smarter, not harder.

mdsohail99:
> As a temp fix lets disable compressed APEX which causes a12 to not boot on 4.14 kernel.
> 
> Please test !!!!!

Hi brother I am noob to developing, can you please tell me or guide me how to do that. Thanks.

ShujathMohd (changes requested):
The flag should be declared immediately/after the inheritance of updatable_apex.mk which is in platform.
By declaring here the flag t is getting overriden/ignored.

bartcubbins:
or we can set it with OVERRIDE_PRODUCT_COMPRESSED_APEX flag

MarijnS95:
`OVERRIDE_PRODUCT_COMPRESSED_APEX` works fine for me, no idea why this wasn't tested.

Haxk20:
Because my pdx201 rebooted once i wiped data. And i didnt get to it yet.
Thus has test please with lot of !

MarijnS95:
> Because my pdx201 rebooted

No need to submit PRs if you know your solution doesn't resolve the issue :)

PR #873, opened 2021-11-11
[s-mr1] overlay: base: Set default value of config_longPressOnPowerBehavior

bartcubbins:
Set default behavior when the user long presses the power button.
This allows to call the power menu with a long press of the power
button. If the default value is set to 5 (Go to assistant), the fallback
to the old behavior will not work when "Hold for Assistant" is disabled.

Reference:
https://android.googlesource.com/platform/packages/apps/Settings/+/34cbca502bdfd6e0e5fff836bb9b0a3abf6811ea

Signed-off-by: Pavel Dubrova <pashadubrova@gmail.com>


MarijnS95:
Thanks, works a treat!

PR #879, opened 2022-01-20
[R] common-treble: use GPT-based IBootControl 1.0 HAL for lena

sledges:
IBootControl 1.1 HAL changed the method of storing boot status information. It now uses the `misc` partition, and requires vendors to ship a compatible bootloader to read from it.
1.0 HAL uses GPT partition entries, and devices with such bootloaders should keep the old HAL.

### DIFF
--- a/common-treble.mk
+++ b/common-treble.mk
@@ -138,8 +138,16 @@ PRODUCT_PACKAGES += \
 
 # Only define bootctrl HAL availability on AB platforms:
 ifeq ($(AB_OTA_UPDATER),true)
+# Define HAL 1.0 for platforms reading it from GPT
+ifneq ($(filter lena, $(SOMC_PLATFORM)),)
+PRODUCT_PACKAGES += \
+    android.hardware.boot@1.0-impl \
+    android.hardware.boot@1.0-impl.recovery \
+    android.hardware.boot@1.0-service
+else
 PRODUCT_PACKAGES += \
     android.hardware.boot@1.1-impl \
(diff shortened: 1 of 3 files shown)


MarijnS95:
> No description provided

Maybe explain the gpt vs misc difference in the 1.0 and 1.1 default hap in a comment and the commit body?

common-treble.mk
```diff
@@ -138,8 +138,16 @@ PRODUCT_PACKAGES += \
 
 # Only define bootctrl HAL availability on AB platforms:
 ifeq ($(AB_OTA_UPDATER),true)
+# Define HAL 1.0 for platforms reading it from GPT
+ifneq ($(filter lena, $(SOMC_PLATFORM)),)
+PRODUCT_PACKAGES += \
+    android.hardware.boot@1.0-impl \
+    android.hardware.boot@1.0-impl.recovery \
+    android.hardware.boot@1.0-service
```
MarijnS95, line 146: I remember that the 1.1 service can start the 1.1 and 1.0 passthrough. Perhaps we only need to select on the -impl for Lena.

Then you can remove the vintf workaround as well.

MarijnS95 (changes requested):

common-treble.mk
MarijnS95: https://android.googlesource.com/platform/hardware/interfaces/+/refs/tags/android-11.0.0_r46/boot/1.1/default/service.cpp#26 fwiw.

sledges:
Superseded by #880.

PR #881, opened 2022-01-31
[R/S] mediacodec: Update seccomp polic from CAF.

Thaodan:
Updated from vendor-qcom-opensource-media as of 182649923d3813e2e539a0e36d83b3e70dcfcc4a:
https://github.com/sonyxperiadev/vendor-qcom-opensource-media/tree/182649923d3813e2e539a0e36d83b3e70dcfcc4a


MarijnS95:
Before doing this, have you seen actual denials?

PR #874, opened 2021-11-25
init: qcrild.rc : Add vendor prefix to ril-daemon

ShujathMohd:
qcrild.rc is attempting to stop ril-daemon which doesnt exist (https://github.com/sonyxperiadev/device-sony-common/blob/s-mr1/rootdir/vendor/etc/init/qcrild.rc#L9) and errors out with the below message.

"init: Command 'stop ril-daemon' action=vendor.rild.libpath=/odm/lib64/libril-qc-hal-qmi.so (/vendor/etc/init/qcrild.rc:9) took 0ms and failed: service ril-daemon not found."

Instead we have vendor.ril-daemon service starting in the background.

"init: starting service 'vendor.ril-daemon'..."

Add the vendor prefix to ril-daemon.

Fixes the buggy sim detection on S.

PR #883, opened 2022-04-13
Revert "init: qcrild.rc : Add vendor prefix to ril-daemon"

sledges:
On R the service is still called "ril-daemon".

This fixes random SIM detection https://github.com/sonyxperiadev/bug_tracker/issues/736.

This reverts commit 30b54ba54619cc17bfc8bfd7a56d8e235579431a.

### DIFF
--- a/rootdir/vendor/etc/init/qcrild.rc
+++ b/rootdir/vendor/etc/init/qcrild.rc
@@ -6,5 +6,5 @@ service vendor.qcrild /odm/bin/hw/qcrild
     capabilities BLOCK_SUSPEND NET_ADMIN NET_RAW
 
 on property:vendor.rild.libpath=/odm/lib64/libril-qc-hal-qmi.so
-    stop vendor.ril-daemon
+    stop ril-daemon
     enable vendor.qcrild


MarijnS95:
Oh and @jerpelea be careful with the rebase script, as this revert will automagically land on `s-mr1`/`master` when rebased on top of `r-mr1` :grimacing:

sledges:
> @sledges We deduced in [#874 (comment)](https://github.com/sonyxperiadev/device-sony-common/pull/874#issuecomment-979126097) that this patch was also applicable to R - can you explain how this is the case? Was that AOSP patch perhaps not picked to the R branch/tag/tree after all?

There is no `/vendor/etc/init/rild.rc` on Sony AOSP filesystem, only `rild.legacy.rc`, within which the service is called `ril-daemon`.

`qcrild.rc` attempts to disable `vendor.ril-daemon`, which fails, and `ps -A | grep ril` returns 3 running daemons, one of them is the legacy one.
Disabling the legacy one properly gets rid of the SIM detection issue.

ix5:
Somehow, a ping about this found its way to my inbox.

I can't be bothered about any of the useless waste of everyone's time that is Google dropping new Android versions, but for my humour and the benefit of the sfos people, here goes:

We had planned to align our "properties" to Treble-compliant naming so that we could set `PRODUCT_COMPATIBLE_PROPERTY` (or rather, `PRODUCT_COMPATIBLE_PROPERTY_OVERRIDE`) to true. Because of a certain company that starts with Q unwilling to fix their insane mess of proprietary RIL-related code, we could not do this and resorted to exemptions granted to devices with old shipping API levels (hence, why SODP [spoofs shipping API levels](https://github.com/sonyxperiadev/device-sony-lena/blob/091b7413ac7fbf869a9d1d8a57d1418c0fa75b15/platform.mk#L93) - I can guarantee that the Xperia 10 Mk. III did not ship with Oreo).

Via some further insanity, the state of "compatible" property names finds its way into the service name of `ril-daemon`:
- When using compatible prop names, the build system will copy over `rild.rc` with the service name set to `vendor.rild-damon`, which I included in the (now reverted) patch.
- When not, it'll copy over `rild.legacy.rc` - see [AOSP: Android.mk](https://cs.android.com/android/platform/superproject/+/android-12.0.0_r31:hardware/ril/rild/Android.mk;l=35-40).

On Android 11, which is now nearing 2 years since pre-release and which SODP and sfos somehow still seem to be stuck on, the exemption for "legacy" props based on shipping API or override is still present:
https://cs.android.com/android/platform/superproject/+/android-11.0.0_r5:build/make/core/config.mk;l=673-680

However, on A12 and current `master` releases, it's gone:
https://cs.android.com/android/platform/superproject/+/android-12.0.0_r28:build/make/core/config.mk;l=609-612

So this "solution" will not fly with future (actually current) AOSP releases.

/over and out.

MarijnS95:
> @MarijnS95 this is why there will be a revert of the revert on top s-mr1 and master

I'd reply with a barbecue emoji if there was one. But since there is none, have a cook instead 🧑‍🍳

> There is no `/vendor/etc/init/rild.rc` on Sony AOSP filesystem, only `rild.legacy.rc`, within which the service is called `ril-daemon`.

@sledges that makes it sound like https://github.com/sonyxperiadev/device-sony-common/pull/874 was incomplete and incorrect at best, or did we fix this for S only?

@ix5 Thanks for your insights as always!

(re. shipping API level spoofing, I don't think the "shipping API" of stock has ~much~ any meaning for SODP at all since it "ships" completely different software and really only maintains firmware and a few random blobs in the process)

jerpelea:
@MarijnS95 @sledges the patch has been applied to s-mr1 and master then reverted to prevent future issues

MarijnS95:
> It shouldn't have been applied to R, but back then both branches shared the same git tree, and without the knowledge about `rild.rc` vs `rild.legacy.rc` that @ix5 has kindly explained, the split had been averted (until now:).

In other words we simply misjudged the patch as I haven't even clicked open any of the diffs to see `PRODUCT_COMPATIBLE_PROPERTY` nor the fallback `rild.legacy.rc` being added, so the suggestion to backport it to R was incorrect _even if_ the upstream patch landed in the R releases.

> @MarijnS95 @sledges the patch has been applied to s-mr1 and master then reverted to prevent future issues

Don't think we should've made it land on R in the first place, but rebasing it off at this stage may only be more confusing history-wise.

PR #884, opened 2022-04-24
Add modem config for Turkcell (TR)

farukr1:
After adding this, I can see the notification for the correct config file however, IMS still displays Not Registered. The same mbn file can be set active using Qualcomm PDC tool and IMS, VoLTE works that way, I'm not sure what the PDC tool does different than the ModemConfigService.

### DIFF
--- a/rootdir/vendor/oem/modem-config/S236.1/modem.conf
+++ b/rootdir/vendor/oem/modem-config/S236.1/modem.conf
@@ -1,0 +1,1 @@
+mcfg_sw/generic/euro/turkcell/vlvw/tr/mcfg_sw.mbn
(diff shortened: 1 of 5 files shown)


MarijnS95:
Are there any more modem configs / cell providers available on the ROM you pulled this from that are not yet in this repository? If so please add those too, that may be useful for others :)

MarijnS95:
Would you be able to squash the commits, and document in the title/body what device and ROM version they originated from? Helps us understand what's there and what's not.

MarijnS95:
Squashing does not mean closing :wink: - you'll have to force-push.

farukr1:
Sorry, I was struggling .'(

MarijnS95:
@farukr1 Performing good contributions is tricky but you're learning fast and nailed it in #885!

farukr1:
@MarijnS95 Thank you for the motivation, I admit I didn't study git well 😅

PR #885, opened 2022-04-24
Add modem config files for Turkish providers

farukr1:
This adds modem config files for the following providers,

- Turkcell
- Turk Telekom
- Vodafone

Firmware files are extracted from latest stock ROM (52.1.A.3.49) at /vendor/firmware_mnt/image/modem_pr/mcfg/configs

### DIFF
--- a/rootdir/vendor/oem/modem-config/S236.1/modem.conf
+++ b/rootdir/vendor/oem/modem-config/S236.1/modem.conf
@@ -1,0 +1,1 @@
+mcfg_sw/generic/euro/turkcell/vlvw/tr/mcfg_sw.mbn
--- a/rootdir/vendor/oem/modem-config/S236.2/modem.conf
+++ b/rootdir/vendor/oem/modem-config/S236.2/modem.conf
@@ -1,0 +1,1 @@
+mcfg_sw/generic/euro/turkcell/vlvw/tr/mcfg_sw.mbn
--- a/rootdir/vendor/oem/modem-config/S237.3/modem.conf
+++ b/rootdir/vendor/oem/modem-config/S237.3/modem.conf
@@ -1,0 +1,1 @@
+mcfg_sw/generic/euro/vodafone/vlvw/tr/mcfg_sw.mbn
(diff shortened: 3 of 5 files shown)


MarijnS95:
Add that information to the _commit_ message as you did with https://github.com/sonyxperiadev/device-sony-common/pull/884/commits/8f9bc13ef6d71182e9e98c7fcfa548bf46e01e21, information in the PR body gets lost from the merged commit.

farukr1:
It should be okay now.

MarijnS95 (approved):
That's almost perfect, thanks!

I couldn't help but notice that 3 out of 5 files are terminated with a newline. As far as I'm aware that shouldn't be the case, can you remove them to prevent possible invalid file paths? I don't think `ModemConfig` strips/trims the path so this'll result in a "file not found".

MarijnS95 (changes requested):
As such I should probably set the review to "changes requested" until this addressed, in hopes to avoid accidental ninja-merges :sweat_smile:

farukr1:
> That's almost perfect, thanks!
> 
> I couldn't help but notice that 3 out of 5 files are terminated with a newline. As far as I'm aware that shouldn't be the case, can you remove them to prevent possible invalid file paths? I don't think `ModemConfig` strips/trims the path so this'll result in a "file not found".

Sorry about that, I'll trim those and update here in the following hours.

farukr1:
Okay, no more sneaky newlines at the end.

MarijnS95:
Almost great! Note that when squashing, you don't need to add things like:

> also: Remove one byte at the end cuz it is whitespace

Because that change is not part of your commit anymore, after squashing them together. As such, future readers will not understand what byte you are talking about; unless this is a remnant from the stock ROM in which case you should clearly mention that, ie. ``Also remove trailing newline in some files respective to stock, to prevent accidentally tripping up SODPs `ModemConfig`.``. And not in the least because shorthands like `cuz` are not very professional...

farukr1:
Sorry about that, it wasn't a remnant from stock so I deleted that line all together.

MarijnS95 (approved):

PR #886, opened 2022-04-25
Replace sm8250 with $(KONA)

LogicFan:
Not all custom ROMs use `KONA = sm8250`. For example, the lineage uses `KONA = kona`. It'll cause problems to hard-code `sm8250` everywhere.

This PR replaces all hard-coded sm8250 that is used for detecting `TARGET_BOARD_PLATFORM` to `$(KONA)`.

### DIFF
--- a/hardware/camera/Android.mk
+++ b/hardware/camera/Android.mk
@@ -3,7 +3,7 @@
 
 LOCAL_PATH := $(call my-dir)
 
-ifneq ($(filter sdm845 sm6350 sm8150 sm8250,$(TARGET_BOARD_PLATFORM)),)
+ifneq ($(filter sdm845 sm6350 sm8150 $(KONA),$(TARGET_BOARD_PLATFORM)),)
 
 include $(SONY_CLEAR_VARS)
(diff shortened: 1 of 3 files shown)


MarijnS95:
I doubt Lineage uses this repository though, they have always had their own device trees which were fully stock-blob based.

LogicFan:
lineage don't because currently, all officially supported devices are using stock-blob. But that doesn't stop someone from making lineage based on sodp (like me 😋). This will just make thing a little bit easier for people who want to make lineage based on sodp for edo platform.

MarijnS95:
Why don't we do the same for `LITO`/`MSMNILE`/`TRINKET`/`LAHAINA` (bar the ninja-merge)?

jerpelea:
@MarijnS95 can be done for all

Paulbouchara:
This would be super useful for custom rom porting on lineage based os

jerpelea:
please PR

MarijnS95:
@jerpelea Done in https://github.com/sonyxperiadev/device-sony-common/pull/888.

PR #892, opened 2022-07-29
[S] hardware/health: Wait for adsp and `power_supply/battery` before starting health HAL

MarijnS95:
On at least our Sagami class of devices the kernel battery driver (including SoMC charger extensions) probes/registers later than the start of our `health` HAL, resulting in it not seeing the battery until a restart.  The power supply arrives as soon as adsp is booted, which is our signal to kick-start the HAL.

Note that we could also have a `wait /sys/class/power_supply/battery` on another `on <xxx>` trigger that starts _later_ than adspstart, but this'll block the entire trigger until a power supply arrives.


MarijnS95:
WARNING: This is untested on other devices!

PR #893, opened 2022-08-18
Updates for Android 13/T

ShujathMohd:


common-packages.mk
```diff
@@ -22,9 +22,7 @@ PRODUCT_PACKAGES += \
 
 # Audio
 PRODUCT_PACKAGES += \
-    audio.a2dp.default \
```
MarijnS95, line 25: If I remember correctly this also needs cleanup in device trees.

MarijnS95 (changes requested):

PR #895, opened 2022-08-31
init: Make bindmount consistent with software binary image

hellobbn:
Signed-off-by: Luofan Chen <clfbbn@gmail.com>


hellobbn:
Please not that the following files are not in the correct location:
`/odm/system_ext/app/datastatusnotification.apk` should be in `/odm/system_ext/app/datastatusnotification/datastatusnotification.apk`
`/odm/system_ext/priv-app/ims.apk` should be in `/odm/system_ext/priv-app/ims/ims.apk`

PR #896, opened 2022-09-01
common-treble.mk: Build hostapd libraries

hellobbn:
Since wpa-suppliant is now switched to AIDL, it no longer
builds the HIDL interfaces that the prebuilt software binaries
need.

Build the HIDL interface to make the prebuilt executable happy.

### DIFF
--- a/common-treble.mk
+++ b/common-treble.mk
@@ -71,7 +71,8 @@ endif
 
 # Wi-Fi
 PRODUCT_PACKAGES += \
-    android.hardware.wifi@1.0-service
+    android.hardware.wifi@1.0-service \
+    android.hardware.wifi.hostapd@1.0.vendor
 
 # NFC packages
 PRODUCT_PACKAGES += \


MarijnS95:
Why isn't this already available on system?

hellobbn:
I am not sure why it is not built by default. And is causing ld errors. Meybe the wifi part is switching to aidl and no longer depends on this? I will check later

MarijnS95:
Add the result of that investigation to the _commit_ message and pull request description, makes it a whole lot easier to review end refer to in hindsight.

hellobbn:
I have updated the commit message and pull request description

MarijnS95:
Thanks for documenting what you're doing here, that allows us/me to properly review this.

I don't think is a workaround we should be making, as you clarify yourself the Android system (since FCM 7) now needs/uses the AIDL version meaning `hostapd` won't work if our vendor compartment hosts a HIDL version instead. But at least we'll get told about this when bumping to FCM 7. It's an optional interface at least.

jerpelea:
our next delivery is using AIDL and the change is valid

MarijnS95:
> our next delivery is using AIDL and the change is valid

No, if our (next) delivery uses AIDL we should omit/remove the HIDL library entirely.

PR #906, opened 2023-10-29
Bluetooth Fixes[#2] 

ShujathMohd:


ShujathMohd:
In one of the commit from this PR we are moving from 

    android.hardware.bluetooth@1.0-service-qti to     android.hardware.bluetooth@1.0-service

android.hardware.bluetooth@1.0-service-qti  was imported to opendevices around android pie I believe.

 android.hardware.bluetooth@1.0-service : https://android.googlesource.com/platform/hardware/interfaces/+/refs/tags/android-14.0.0_r10/bluetooth/1.0/default/android.hardware.bluetooth%401.0-service.rc

 android.hardware.bluetooth@1.0-service-qti  : https://github.com/sonyxperiadev/vendor-qcom-opensource-bluetooth/blob/master/interfaces/bluetooth/1.0/qti/android.hardware.bluetooth%401.0-service-qti.rc


The only change I notice is https://github.com/sonyxperiadev/vendor-qcom-opensource-bluetooth/blob/master/interfaces/bluetooth/1.0/qti/android.hardware.bluetooth%401.0-service-qti.rc#L5

However during the testing did not come across any sleep or wakelock issues while being connected to bluetooth devices.

jerpelea (changes requested):
BT is platform dependent and should be kept separate

jerpelea (approved):

jerpelea (approved):

MarijnS95 (approved):
Changes seem okay, did you confirm that Bluetooth Audio still uses the offload path?

The commit messages could use elaboration though, as pointed out in other reviews.

> `E audiohalservice: Failed to dlopen android.hardware.bluetooth.audio-impl.so: dlopen failed: library "android.hardware.bluetooth.audio-impl.so" not found`

Isn't it falling back to the non-AIDL implementation, or do we have the wrong passthrough loader (or is it the audio service itself that loads the `-impl` nowadays)?

PR #910, opened 2024-01-08
[u-mr1] common-prop: Set ro.hardware.egl property

bartcubbins:
A recent AOSP change[1] has put a hard dependency on ro.hardware.egl or equivalent property being set, otherwise it fails to load EGL libraries successfully:
D libEGL  : Failed to load drivers from property ro.board.platform
            with value sm8450
F libEGL  : couldn't find an OpenGL ES implementation, make sure
            one of persist.graphics.egl, ro.hardware.egl and
            ro.board.platform is set
--------- beginning of crash
F DEBUG   : Cmdline: /system/bin/surfaceflinger
F DEBUG   : pid: 573, tid: 600, name: surfaceflinger  >>> /system/bin/surfaceflinger <<<
F DEBUG   : uid: 1000
F DEBUG   : tagged_addr_ctrl: 0000000000000001 (PR_TAGGED_ADDR_ENABLE)
F DEBUG   : signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
F DEBUG   : Abort message: 'couldn't find an OpenGL ES implementation,
            make sure one of persist.graphics.egl, ro.hardware.egl and
            ro.board.platform is set'

[1] https://android.googlesource.com/platform/frameworks/native/+/8968823ade16061d589c1f9f7562a1de653be2b1

### DIFF
--- a/common-prop.mk
+++ b/common-prop.mk
@@ -297,6 +297,10 @@ PRODUCT_PROPERTY_OVERRIDES += \
 PRODUCT_PROPERTY_OVERRIDES += \
     ro.crypto.scrypt_params=15:3:1
 
+# Look for libGLES_adreno.so instead of libGLES_$(BOARD_TARGET_PLATFORM).so
+PRODUCT_PROPERTY_OVERRIDES += \
+    ro.hardware.egl=adreno
+
 # Look for vulkan.qcom.so instead of vulkan.$(BOARD_TARGET_PLATFORM).so
 PRODUCT_PROPERTY_OVERRIDES += \
     ro.hardware.vulkan=qcom


jerpelea (approved):

MarijnS95:
Does this mean that we also had a symlink before? If so, can we drop it?

MarijnS95:
I.e. for Vulkan below, Android searches for `$LIBPATH/hw/vulkan.${ro.hardware.vulkan}.so`.

What lib will EGL now search for, considering e.g. https://github.com/sonyxperiadev/device-sony-common/blob/84a56b608157a69bdc1d700e52d6f49d3121016e/hardware/adreno/Android.mk?

bartcubbins:
> Does this mean that we also had a symlink before? If so, can we drop it?

Unfortunately we can't https://android.googlesource.com/platform/frameworks/native/+/refs/tags/android-14.0.0_r21/opengl/libs/EGL/Loader.cpp#437



> I.e. for Vulkan below, Android searches for `$LIBPATH/hw/vulkan.${ro.hardware.vulkan}.so`.
> 
> What lib will EGL now search for, considering e.g. https://github.com/sonyxperiadev/device-sony-common/blob/84a56b608157a69bdc1d700e52d6f49d3121016e/hardware/adreno/Android.mk?

Vulkan library loader is located in a different location and is not associated with the EGL https://android.googlesource.com/platform/frameworks/native/+/refs/tags/android-14.0.0_r21/vulkan/libvulkan/driver.cpp#168

But it would be nice if `ro.hardware.vulkan == ro.hardware.egl`

MarijnS95:
They're obviously different libraries and it's fine IMO to have different props for Vulkan and EGL. I was just wondering how Android was finding `libEGL_adreno.so` before this change (without looking at the diff) and assuming we used a different message, e.g. a symlink?

bartcubbins:
wildcard:) libEGL_*.so the first lib that was encountered was loaded:)
P.S. in the latest BSP QCOM delivers Vulkan lib as vulkan.adreno.so;)

Issue #916, opened 2024-07-04
Unresolved package dependencies for android-14.0.0_r50, Xperia 1 V


gecko19:
Hello All, 

I am new to the Sony AOSP build. I am trying to build the AOSP image for the Xperia 1 V.

I face dependency errors when building branch android-14.0.0_r50 from the sony manifest repo, as seen below.
Any ideas how I could resolve this.

Thanks,
George.

```
[100% 1165/1165] analyzing Android.bp files and generating ninja file at out/soong/build.aosp_xqdq54.ninja
FAILED: out/soong/build.aosp_xqdq54.ninja
cd "$(dirname "out/host/linux-x86/bin/soong_build")" && BUILDER="$PWD/$(basename "out/host/linux-x86/bin/soong_build")" && cd / && env -i  "$BUILDER"     --top "$TOP"     --soong_out "out/soong"     --out "out"     --soong_variables out/soong/soong.aosp_xqdq54.variables -o out/soong/build.aosp_xqdq54.ninja --globListDir aosp_xqdq54 --globFile out/soong/globs-aosp_xqdq54.ninja -l out/.module_paths/Android.bp.list --available_env out/soong/soong.environment.available --used_env out/soong/soong.environment.used.aosp_xqdq54.build Android.bp
error: device/sony/common/hardware/health/Android.bp:1:1: "android.hardware.health@2.0-service.sony" depends on undefined module "android.hardware.health@2.0-impl".
Or did you mean ["android.hardware.health@2.1-impl"]?
error: device/sony/common/hardware/health/Android.bp:1:1: "android.hardware.health@2.0-service.sony" depends on undefined module "libhealthservice".
Or did you mean ["libartservice"]?
error: device/sony/common/hardware/health/Android.bp:1:1: "android.hardware.health@2.0-service.sony" depends on undefined module "libhealthstoragedefault".
Or did you mean ["healthfitness-ui-statsd"]?
fatal errors encountered
00:03:51 soong bootstrap failed with: exit status 1

```

raffles496:
Hi friend.
I have recently started building AOSP Android 14 for my Xperia 10 IV, using the repositories and instructions provided by Sony. Unfortunately I have encountered the same problem that you describe.
Although I have investigated the reason for the error, I have not found anything concrete.

```
[100% 1/1] analyzing Android.bp files and generating ninja file at out/soong/build.aosp_xqcc54.ninja
[...]

PR #917, opened 2024-07-05
Migrate to health hal 2.1

ShujathMohd:
Starting U QPR3 health hal 2.0 is deprecated.


ShujathMohd:
Fixes #916

MarijnS95:
So is `health 2.1`.

I suggest you go through the _default implementation_ of `health 2.1` that you're using here, and document why you think this is the correct, complete replacement of Sony-specific code that is being deleted in this PR.

PR #920, opened 2024-08-22
[t-mr1] Inital Audioreach customization

bartcubbins:
Implement changes required for Audioreach audio subsystem.

1. Bump audio implementation to 7.1 (All current supported platforms should receive audio
policy updates in accordance with[1])
2. Bump audio effect implementation to 7.0.
3. Bump soundtrigger implementation to 2.3.
4. Switch to 64bit audio service.

Changes tested on at least one device from each supported platform except Lena.

[1] https://source.android.com/docs/core/audio/hidl-implement#xml-schema-hal7

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -81,6 +81,11 @@ SOONG_CONFIG_NAMESPACES += qti_kernel_headers
 SOONG_CONFIG_qti_kernel_headers := version
 SOONG_CONFIG_qti_kernel_headers_version := $(SOMC_KERNEL_VERSION)
 
+# Build 64bit audio service
+SOONG_CONFIG_NAMESPACES += android_hardware_audio
+SOONG_CONFIG_android_hardware_audio := run_64bit
+SOONG_CONFIG_android_hardware_audio_run_64bit := true
+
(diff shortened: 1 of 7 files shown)


common.mk
```diff
@@ -81,6 +86,11 @@ SOONG_CONFIG_NAMESPACES += qti_kernel_headers
 SOONG_CONFIG_qti_kernel_headers := version
 SOONG_CONFIG_qti_kernel_headers_version := $(SOMC_KERNEL_VERSION)
 
+# Build 64bit audio service
+SOONG_CONFIG_NAMESPACES += android_hardware_audio
+SOONG_CONFIG_android_hardware_audio := run_64bit
+SOONG_CONFIG_android_hardware_audio_run_64bit := true
```
MarijnS95, line 92: Do we also need `AUDIOSERVER_MULTILIB := first`? I used to have that when enabling 64-bit audio on my local device (but it was only used by soundtrigger iirc).

bartcubbins: It's finally not a thing anymore

PR #921, opened 2024-08-22
[t-mr1] etc: init.common.rc: Restart Health HAL when ADSP ready

bartcubbins:
Restart Health HAL when ADSP is out of reset.
On devices where communication with the battery occurs over GLINK,
it is important that Health HAL is (re)launched after ADSP ready.

This is reimplementation of https://github.com/sonyxperiadev/device-sony-common/commit/7814953e62a8b4f48e51fe8f3ddebbbca51aebd5 ("hardware/health: Wait for
adsp and power_supply/battery before starting health HAL") which
was removed in https://github.com/sonyxperiadev/device-sony-common/commit/57da52d8d8bfdf02a8506edcb276fe2fe0657172 ("hardware/health: Upgrade to 2.1 via
libhealth2impl").

This brings back the working battery status on platforms such as Sagami, Nagara and Yodo.

### DIFF
--- a/hardware/health/CycleCountBackupRestore.cpp
+++ b/hardware/health/CycleCountBackupRestore.cpp
@@ -35,6 +35,13 @@ namespace health {
 static constexpr int kCCBackupTrigger = 20;
 
 CycleCountBackupRestore::CycleCountBackupRestore() {
+    kPersistCycleFile = "/mnt/vendor/persist/battery/battery_cycle_count";
+    kSysCycleFile = "/sys/class/power_supply/bms/cycle_count";
+
+    if (access(kSysCycleFile.c_str(), F_OK)) {
+        kSysCycleFile = "/sys/class/power_supply/battery/cycle_count";
+    }
+
     persist_cycles = 0;
     sysfs_cycles = 0;
     saved_soc_ = -1;
--- a/hardware/health/CycleCountBackupRestore.h
+++ b/hardware/health/CycleCountBackupRestore.h
@@ -22,9 +22,6 @@ namespace device {
 namespace sony {
 namespace health {
 
-static const std::string kSysCycleFile = "/sys/class/power_supply/bms/cycle_count";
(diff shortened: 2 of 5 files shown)


MarijnS95:
So much for:

> [!CAUTION]
> This pull request has **only been** build-tested on Android T, and has **never been tried** on an actual device yet!
>
> Please help me do so if you have the opportunity.

:joy: 

Thanks for actually doing so. Bit unfortunate that we cannot leave the HAL `disabled` by default as was the case before... Unless we can have an `on init` or other trigger that first calls `disable`?

MarijnS95:
Do we also miss the `chown` permissions?

jerpelea (approved):

bartcubbins:
The path has changed from `/sys/class/power_supply/bms/` to `/sys/class/power_supply/battery/`. It seems only pre-Edo platforms use the old path, I'll check all platforms manually and let you know

bartcubbins:
| /sys/class/power_supply/bms/  | /sys/class/power_supply/battery/ |
| ------------- | ------------- |
| Tama | Murray  |
| Kumano | Sagami  |
| Seine | Nagara  |
| Lena | Yodo  |
| Zambezi |
| Edo |

Based on this I will add a simple check for the existence of the `bms` path to the Health HAL and if it does not exist, we make a fallback to the `battery` path. I will also add permissions for both paths

bartcubbins:
| /sys/class/power_supply/bms/  | /sys/class/power_supply/battery/ |
| ------------- | ------------- |
| Tama | Murray  |
| Kumano | Sagami  |
| Seine | Nagara  |
| Lena | Yodo  |
| Zambezi | (upcoming) Asahi |
| Edo | (upcoming) Columbia |

Based on this I will add a simple check for the existence of the `bms` path to the Health HAL and if it does not exist, we make a fallback to the `battery` path. I will also add permissions for both paths

jerpelea (approved):

PR #925, opened 2024-09-13
[r-mr1] Move Wifi config to own RRO

voidanix:
Might have to get into S/T/U/V too


jerpelea (approved):

jerpelea:
we can include it from Android T-mr1

PR #929, opened 2024-11-05
Use parseint instead of stoi

rinigus:
For some reason exception handling was not working for stoi-based battery cycles parsing. Replace with ParseInt

The only concern I have is that it can lead to some other exception in static_cast<> used by ParseInt

Related issue: https://github.com/sonyxperiadev/bug_tracker/issues/850

### DIFF
--- a/hardware/health/LearnedCapacityBackupRestore.cpp
+++ b/hardware/health/LearnedCapacityBackupRestore.cpp
@@ -25,6 +25,7 @@
 #include <android-base/file.h>
 #include <android-base/logging.h>
 #include <android-base/strings.h>
+#include <android-base/parseint.h>
 
 #include "LearnedCapacityBackupRestore.h"
 
@@ -74,12 +75,7 @@ void LearnedCapacityBackupRestore::ReadFromPersistStorage() {
     }
 
     buffer = ::android::base::Trim(buffer);
-    try {
-        persist_capacity = std::stoi(buffer);
-    } catch (std::out_of_range &e) {
-        LOG(WARNING) << "Battery capacity in persist storage file is out of bounds: " << buffer;
-        return;
-    } catch (std::invalid_argument &e) {
+    if (!android::base::ParseInt(buffer.c_str(), &persist_capacity)) {
         LOG(WARNING) << "Data format is wrong in the battery capacity persist file: " << buffer;
         return;
     }
--- a/hardware/health/CycleCountBackupRestore.cpp
+++ b/hardware/health/CycleCountBackupRestore.cpp
@@ -25,6 +25,7 @@
 #include <android-base/file.h>
 #include <android-base/logging.h>
 #include <android-base/strings.h>
+#include <android-base/parseint.h>
 
 #include "CycleCountBackupRestore.h"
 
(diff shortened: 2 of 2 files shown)


MarijnS95:
Did you check if `sscanf("%d")` could/should also use `ParseInt()`, if there's no trailing string?

MarijnS95:
> For some reason exception handling was not working for stoi-based battery cycles parsing. Replace with ParseInt

Probably the most vauge description / reason ever, but I don't have time to help debug this oddity either 😞

hardware/health/LearnedCapacityBackupRestore.cpp
```diff
@@ -74,12 +75,7 @@ void LearnedCapacityBackupRestore::ReadFromPersistStorage() {
     }
 
     buffer = ::android::base::Trim(buffer);
-    try {
-        persist_capacity = std::stoi(buffer);
-    } catch (std::out_of_range &e) {
-        LOG(WARNING) << "Battery capacity in persist storage file is out of bounds: " << buffer;
-        return;
-    } catch (std::invalid_argument &e) {
+    if (!android::base::ParseInt(buffer.c_str(), &persist_capacity)) {
```
MarijnS95, line 78: Note that `Trim()` above goes from the `::android` namespace. Not that it makes too big of a difference.

rinigus: Indeed. I wasn't sure which to use as below there is from `android::base::`

rinigus:
Thanks for looking into it! 

> Did you check if sscanf("%d") could/should also use ParseInt(), if there's no trailing string?

I don't think I follow your question. So, reply is probably "no, I didn't"

Commit message changed

MarijnS95:
> > Did you check if `sscanf("%d")` could/should also use `ParseInt()`, if there's no trailing string?
> 
> I don't think I follow your question. So, reply is probably "no, I didn't"

We found in https://github.com/sonyxperiadev/bug_tracker/issues/850#issuecomment-2453084096 that `LearnedCapacityBackupRestore` did something different from `CycleCountBackupRestore`, namely using `sscanf()` instead of `std::atoi()` and couldn't come to a conclusion why:

https://github.com/sonyxperiadev/device-sony-common/blob/92cc90be2171a32501cb3a43e25c90f5b4b0a0db/hardware/health/LearnedCapacityBackupRestore.cpp#L113

If both are parsing a whole number without any trailing data, both should use `ParseInt` I think.

rinigus:
I am not convinced in changing from `sscanf` to `ParseInt`. `ParseInt` can throw exceptions via its use of casting and exceptions handling seemed to be off in this case for some reason. I don't plan to test what are the differences between those scan methods as I would prefer to focus on other tasks. If you think that in this case we should not merge this PR, we can either keep it open or close it.

MarijnS95:
I don't like that we had an inconsistency here in the first place with no clear documentation/indication why two otherwise-mostly-identical classes use different methods to parse integers out of strings. If there's anything we can do to solve an issue that affects one class but not the other, is to reuse the implementation from the class that doesn't crash and has been battle-tested?

rinigus:
I agree, it would make sense to have the same implementation throughout. Just as almost nothing works on device, its hard to prioritize this. In addition, we still don't know why the original code failed as it is, in my understanding, perfectly in line with what it is expected to be.

rinigus:
I think a better solution has been proposed (https://github.com/sonyxperiadev/bug_tracker/issues/850#issuecomment-2466645744) and I suggest to close this PR.

Issue #931, opened 2024-12-14
 vendor.qti.hardware.display.config.IDisplayConfig compatibility error


itviewer:
```
The following HALs in device manifest are not declared in FCM <= level 5: 
  android.hardware.audio.effect@7.0::IEffectsFactory/default
  android.hardware.audio@7.1::IDevicesFactory/default
  android.hardware.bluetooth.audio.IBluetoothAudioProviderFactory/default (@4)
  android.hardware.cas.IMediaCasService/default (@1)
  android.hardware.security.keymint.IKeyMintDevice/default (@2)
  android.hardware.security.keymint.IRemotelyProvisionedComponent/default (@2)
  android.hardware.security.secureclock.ISecureClock/default (@1)
  android.hardware.security.sharedsecret.ISharedSecret/default (@1)
  android.hardware.wifi.IWifi/default (@2)
  android.hardware.wifi.hostapd.IHostapd/default (@2)
  android.hardware.wifi.supplicant.ISupplicant/default (@3)
  vendor.qti.hardware.display.config.IDisplayConfig/default (@5)
ERROR: files are incompatible: The following instances are in the device manifest but not specified in framework compatibility matrix: 
    vendor.qti.hardware.display.config.IDisplayConfig/default (@5)
Suggested fix:
1. Update deprecated HALs to the latest version.
2. Check for any typos in device manifest or framework compatibility matrices with FCM version >= 5.
3. For new platform HALs, add them to any framework compatibility matrix with FCM version >= 5 where applicable.
4. For device-specific HALs, add to DEVICE_FRAMEWORK_COMPATIBILITY_MATRIX_FILE or DEVICE_PRODUCT_COMPATIBILITY_MATRIX_FILE.: Success
INCOMPATIBLE
21:47:52 ninja failed with: exit status 1
```

jerpelea:
@bartcubbins can you push the fix?

bartcubbins:
@itviewer please specify platform name

PR #932, opened 2024-12-16
[u-mr1] vintf: Add vendor.qti.hardware.display.config v5 to FCM

bartcubbins:
ERROR: files are incompatible: The following instances are in the
device manifest but not specified in framework compatibility matrix:
  vendor.qti.hardware.display.config.IDisplayConfig/default (@5)

Fixes: https://github.com/sonyxperiadev/device-sony-common/issues/931

### DIFF
--- a/vintf/framework_compatibility_matrix.xml
+++ b/vintf/framework_compatibility_matrix.xml
@@ -164,7 +164,7 @@
     </hal>
     <hal format="aidl">
         <name>vendor.qti.hardware.display.config</name>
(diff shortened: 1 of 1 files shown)


jerpelea (approved):

mlehtima:
Was it intentional that this went only to master branch and not also to u-mr1 branch?

jerpelea:
it was pushed on the wrong branch 
I manually re-based it now on all needed branches
thanks

PR #944, opened 2025-02-05
[u-mr1] common: Do not provide generic audio policy volumes

bartcubbins:
Audio policy volumes are tuned for each device individually,
so we cannot provide a generic version. On some devices, the
values may differ from the generic version by a few units,
which is completely unnoticeable, while on others, the values
may vary by 1.5-2 times, which could potentially damage the
speaker.


bartcubbins:
@jerpelea

jerpelea (approved):

PR #960, opened 2026-03-17
[u-mr1] common: Use the UFS BSG framework in gpt-utils for slot switching

bartcubbins:
Use the UFS BSG framework in gpt-utils for slot switching
instead of the UFS IOCTL implementation.

This change should also be applied to the v-mr1 branch.


jerpelea:
please change the target branch to android-16 and it will be auto ported to master

jerpelea (changes requested):
please change target branch

bartcubbins:
@jerpelea

PR #962, opened 2026-03-23
[android-16] Switch to QTI AIDL audio HAL

bartcubbins:
We no longer have any platforms that use the legacy Elite
audio framework. Migrate to the new QTI AIDL audio HAL.
Remove support for the legacy audio and add the common
configs required for the new framework.


bartcubbins:
The PR is in draft until the configs are ported for all platforms.

jerpelea (approved):

